Issue #10, opened 2025-04-14
write code for submitting prompt and compiling actual prompts


mnaydan:

laurejt:
The AI Sandbox does not appear to support the responses API at this time.

PR #23, opened 2025-04-18
Experiments/sandbox utils

laurejt:
ref: #10 

### Key Changes
- Add basic AI sandbox utilities to create a client and run a basic text prompt
- Updated README with instructions on setting up environment with the API sandbox key

### Notes
- marimo has been configured so that it knows about the `remarx` package and so it can be directly imported
- Make sure to set up your local `.env` file

### Evaluation

- [x] Check that the sandbox utility methods can be called from a marimo notebook and run as expected

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,15 @@ pre-commit install
 This branch contains code and data for preliminary experiments with
 Princeton's AI Sandbox.
 
+### AI Sandbox API key setup
+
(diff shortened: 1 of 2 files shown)


rlskoeser (approved):
Works. Was able to import into a test notebook, submit a prompt, and get a response back.

A shorter method name might be nice (maybe `submit_prompt` ?), but doesn't matter too much at this stage.

Issue #8, opened 2025-04-14
at least one prompt template in repo for title mentions


mnaydan:
- store prompts in prompts folder of repo

PR #25, opened 2025-04-21
Experiments/title mentions notebook

laurejt:
**Associated Issue(s):** #10, #8 

### Changes in this PR

- Updated AI Sandbox utilities to use latest API version
- Added basic (zero-shot) and one-shot title mention prompts
- **Added notebook utility for ~displaying annotated responses~ highlighting bracketed text**
- Added notebook exploring title mentions prompts
- **Added AI Sandbox utility for extracting model response text**
- **Added notebook utility for validating bracketed expressions**

### Notes

- Notebook requires no configuration changes beyond what is already required (i.e., `.env` file is properly defined)

### Evaluation

- [x] Check that templates look reasonable.
- [x] Check that notebook runs locally without errors.
- [x] Check that notebook's prose is legible and sufficient.
- [x] **Check that updated notebook runs locally without error.**

### DIFF
--- a/notebooks/title_mentions.py
+++ b/notebooks/title_mentions.py
@@ -1,15 +1,15 @@
 import marimo
 
 __generated_with = "0.12.10"
-app = marimo.App()
+app = marimo.App(width="medium")
 
 
 @app.cell
 def _():
     import marimo as mo
-    from remarx.sandbox_utils import submit_prompt
-    from remarx.notebook_utils import display_bracketed_response
-    return display_bracketed_response, mo, submit_prompt
+    from remarx.sandbox_utils import submit_prompt, get_text_response
+    from remarx.notebook_utils import highlight_bracketed_text
+    return get_text_response, highlight_bracketed_text, mo, submit_prompt
 
 
 @app.cell
@@ -20,7 +20,7 @@ def _(mo):
 
 @app.cell(hide_code=True)
 def _(mo):
-    mo.md(r"""Below is a sample page of text with several title mentions including _Das Kapital_ and _The Communist Manifesto_.""")
+    mo.md(r"""Below is a sample page of text with several title mentions including _Das Kapital_ and _The Communist Manifesto_ with title mentions (with volume information) highlighted.""")
     return
 
 
@@ -197,8 +197,8 @@ def _(basic_response):
 
 
 @app.cell
-def _(basic_response, display_bracketed_response):
-    display_bracketed_response(basic_response)
+def _(basic_response, get_text_response, highlight_bracketed_text):
+    highlight_bracketed_text(get_text_response(basic_response))
     return
 
 
@@ -240,10 +240,10 @@ def _(one_shot_response):
 
 
 @app.cell
-def _(display_bracketed_response, one_shot_response):
+def _(get_text_response, highlight_bracketed_text, one_shot_response):
     # Finds the texts of interest but does not include volume-level information.
     # Misses the additional speech (https://www.marxists.org/deutsch/archiv/marx-engels/1864/10/inaugadr.htm) in the first paragraph
-    display_bracketed_response(one_shot_response)
+    highlight_bracketed_text(get_text_response(one_shot_response))
     return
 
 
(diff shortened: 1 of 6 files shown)


src/remarx/notebook_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    finish_reason = output_object.finish_reason
+    if finish_reason == "length":
+        print("WARNING: Response stopped early due to length limits!")
+    elif finish_reason == "content_filter":
+        print("WARNING: Content filtering occurred for response!")
+
+    # Display reponse
+    output_text = output_object.message.content
+    output_text = output_text.replace("[", "<mark>")
+    output_text = output_text.replace("]", "</mark>")
```
rlskoeser, line 27: Nice. Did you run into cases where the text include either bracket? Any danger of these not matching? (Tolerable for this phase, but curious)

notebooks/title_mentions.py
```diff
@@ -0,0 +1,322 @@
...
+    mo.md(
+        r"""
+        ## Conclusions
+        - Volume-level information is generally ignored.
+        - Prompting to use square brackets for annotations appears to work, will be relatively easy to convert for viewing assuming the underlying text does not include square brackets.
+        - Not clear that adding an example to the prompt helps.
+        - The newer prompt template appears to take a lot more time, but it seems to be relative to the length of the input passage.
+        - Annotations do not appear to be particularly stable between runs.
+            - Spurious annotations may appear in passages without any title references
+            - The speech written by Karl Marx is not necessarily identified
```
rlskoeser, line 315: Did you mean text here?
```suggestion
            - The text written by Karl Marx is not necessarily identified
```

notebooks/title_mentions.py
```diff
@@ -0,0 +1,322 @@
+import marimo
+
+__generated_with = "0.12.10"
+app = marimo.App()
+
+
+@app.cell
+def _():
+    # Notebook for exploring title mention prompts
+    return
```
rlskoeser, line 10: Did you not make it markdown because of the import order?

notebooks/title_mentions.py
```diff
@@ -0,0 +1,322 @@
...
+
+
+@app.cell
+def _():
+    import marimo as mo
+    from remarx.sandbox_utils import submit_prompt
+    from remarx.notebook_utils import display_bracketed_response
+    return display_bracketed_response, mo, submit_prompt
```
rlskoeser, line 18: When I opened the python file in my editor, the ruff fix looked like this:
```python
    import marimo as mo

    from remarx.notebook_utils import display_bracketed_response
    from remarx.sandbox_utils import submit_prompt
    return display_bracketed_response, mo, submit_prompt
```

But even with that change it still gets flagged in the marimo editor 🤷

rlskoeser (approved):
Was able to run locally.  Seems clear enough. It is _very_ slow! Any idea what makes it so slow? Seems slower than what I was testing, but I guess I never ran very many at once.

notebooks/title_mentions.py
laurejt: It's a speech, an [inaugural address](https://www.marxists.org/deutsch/archiv/marx-engels/1864/10/inaugadr.htm)

src/remarx/notebook_utils.py
laurejt: I hadn't, but that's a good point. One thing I've been considering is that perhaps this function should just take a string as input rather than the message object itself. Then, I could do more checks.

notebooks/title_mentions.py
laurejt: I see this kind of a warning a side-effect of the experimental nature of the LSP feature.

notebooks/title_mentions.py
laurejt: I reordered the cells, it appears that marimo requires the imports to go first?

src/remarx/notebook_utils.py
rlskoeser: I also wondered about splitting the error handling from the text munging, they seem like separate functions but I didn't comment because this might be good enough for the current phase.

notebooks/title_mentions.py
rlskoeser: I wondered about that - makes sense that you can't output until you've imported marimo

rlskoeser (approved):
Confirmed it runs locally without errors.

Issue #9, opened 2025-04-14
at least one prompt template in repo for direct quote finding


mnaydan:
- store prompts in prompt folder in repo

rlskoeser:
Added basic and one-shot prompts based on the title mentions prompts. Included in #26

PR #26, opened 2025-04-22
Quotation prompts and demo notebook

rlskoeser:
**Associated Issue:** #9 

### Changes in this PR

- Add prompt directory and template files for quotation prompt (basic and one shot), based on title mentions
- Add quotation prompting demo notebook

### Notes

- Prompts have been written using the bracket style annotation; already ran into a page with a `[1]` in the text but decided to ignore it and move forward for now
- Notebook is using two pages of text from the quotation annotation subset data; one with a long passage and the other with two annotations that directly follow each other
- Notebook uses existing methods for submitting a prompt, processing the response, and highlighting the model annotation
- Only ran the prompts against the default model

### Evaluation

- [ ] Check that templates look reasonable.
- [ ] Check that notebook runs locally without errors.
- [ ] Check that notebook's prose is legible and sufficient.


notebooks/quotation-prompting.py
```diff
@@ -0,0 +1,248 @@
...
+
+@app.cell
+def _(basic_responses, get_text_response, highlight_bracketed_text):
+    highlight_bracketed_text(get_text_response(basic_responses[0]))
+    return
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""Substantial overlap with the annotated passage; it starts about a sentence later and does not include the citation.""")
```
laurejt, line 187: Might be worth adding it includes the full quotation, just not the preceding text within our manual annotation.

notebooks/quotation-prompting.py
```diff
@@ -0,0 +1,248 @@
...
+
+@app.cell
+def _(basic_responses, get_text_response, highlight_bracketed_text):
+    highlight_bracketed_text(get_text_response(basic_responses[1]))
+    return
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""Matches the annotated text. Because of the way we're highlighting, I'm not sure if this is identified as one long quotation or two, as in the annotated data and as indicated by the quotes.""")
```
laurejt, line 199: This should be verifiable by counting how many square brackets are in the text response

notebooks/quotation-prompting.py
```diff
@@ -0,0 +1,248 @@
...
+
+@app.cell
+def _(get_text_response, highlight_bracketed_text, one_shot_responses):
+    highlight_bracketed_text(get_text_response(one_shot_responses[1]))
+    return
+
+
+@app.cell
+def _(mo):
+    mo.md(r"""The results from the one-shot prompt are exactly the same as the zero-shot.""")
```
laurejt, line 243: At can at least confirm this is not necessarily the case on repeated runs. When I ran it, the one-shot included spurious spans for both examples (including what appears to be quoted concepts as well as title references)

notebooks/__marimo__/quotation-prompting.ipynb
laurejt, line 1: Unfortunately it looks like the text highlighting does not show up in this notebook. Seems to be more of a case for saving an HTML copy instead.

laurejt (approved):

Issue #13, opened 2025-04-14
compare annotation output and AI-generated output


mnaydan:
- method and notebook(s) to review/compare annotation + ai output
- create provisional data to use for review so we're not hitting the api every time
  - can we include title of the source somewhere? (how?)
  - for titles, can use the google doc examples testing out templates as well as what's in the google sheet (with bolded title mentions)

PR #27, opened 2025-04-22
Methods & styles to review & compare annotations

rlskoeser:
**Associated Issue(s):** #13

### Changes in this PR

- Adds several methods to notebook_utils for comparing and reviewing annotations
   - side-by-side comparison (two texts)
   - overlap/overlay comparison (two or more)
   - method to check if texts differ (disregarding brackets)
   - method to generate an HTML diff if texts differ
- Adds custom css for highlighting; required for comparison methods to display properly
- Updates the quotation-prompting notebook to use the new comparison methods

### Notes

- This PR is branched from #26 ; that PR should be reviewed first
- I'm working in dark mode; I expect the html diff to look better in light mode than it does in dark mode. I hope the colors I picked are visible enough to work for both light and dark modes, but if not maybe we change colors and force this notebook to be light mode (assuming that's possible)

### Evaluation

- [x] Check that the notebook runs without errors
- [x] Check that highlighted `<mark>` spans are highlighted with a underline color instead of default yellow background color
- [x] Check that side-by-side comparison displays
- [x] Check that html diff is displayed when the models return a modified version of the original text
- [x] Check that the overlapping comparison display works well enough (does not work well when texts are significantly difficult)
- [x] Check that the highlight colors are visible enough


laurejt (approved):
Looks good enough to merge. In the future, we will need to modify the display format for highlighting annotations. The current colors are not ideal for light mode and they are not color-blind friendly.

Issue #11, opened 2025-04-14
run the prompts through the models


mnaydan:
- store output (csv) in repo

optional / stretch:
- run prompts on untagged text to see how many references we get where we expect none

PR #30, opened 2025-04-23
Quotation prompt model run script & data

rlskoeser:
**Associated Issue(s):** #11

### Changes in this PR

- Created `get_quotations.py` - simplified version of `src/remarx/get_title_mention_responses.py` script
- Added an option to pass in and reuse an api client when submitting prompts
- Renamed direct quotation subset to match title subset path/name
- Data file of quotations generated by the script from three models

### Notes

- If you want to run the script locally:
```console
python src/remarx/get_quotations.py data/direct_quotes_subset.csv prompts/quotations/basic.txt \
data/model_responses/quotations_test.csv
```

- The data file was manually combined from separate runs since I had a bug in the row data; I did finally correct the bug. It may include multiple responses from the same model for the same annotation id; thought that could be interesting in case it varied (haven't checked if they are exact dupes but wouldn't be too hard).
- Thought about adding a max number arg so you could run on a subset of fewer records but didn't want to increase complexity

### Evaluation

- [x] Review the quotation data output and confirm it looks reasonable
- [x] Confirm you can start the script without error (don't think you need to do a complete run)


src/remarx/sandbox_utils.py
```diff
@@ -35,16 +35,15 @@ def create_client() -> AzureOpenAI:
 
 
 def submit_prompt(
-    task_prompt,
-    user_prompt,
-    model="gpt-4o",
+    task_prompt, user_prompt, model="gpt-4o", client=None
 ) -> ChatCompletion:
```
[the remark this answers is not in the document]
rlskoeser: I'll update and resolve after your PR is merged.

rlskoeser:
@laurejt can't believe I forgot to add the data! Added that and made the other corrections you flagged. Should we get yours merged first and then I'll update mine to reconcile differences?

data/model_responses/quotations.csv
laurejt, line 1: Should there be any results for `o3-mini`?

src/remarx/sandbox_utils.py
```diff
@@ -35,7 +36,7 @@ def create_client() -> AzureOpenAI:
 
 
 def submit_prompt(
-    task_prompt, user_prompt, model="gpt-4o", client=None
+    task_prompt, user_prompt, model="gpt-4o", client: Optional[AzureOpenAI] = None
```
laurejt, line 39: Happy to use `Optional` but also open to this instead

```suggestion
    task_prompt, user_prompt, model="gpt-4o", client: None | AzureOpenAI = None
```

data/model_responses/quotations.csv
rlskoeser: There should, thanks for catching. I've added them into the combined set.

PR #31, opened 2025-04-23
Experiments/view title mentions results

laurejt:
**Associated Issue(s):** 29

### Changes in this PR

- Adds a new notebook for exploring the title mentions model response data

### Notes

- Notebook relies on the following two files:
  1. `data/title_mentions_subset.csv`
  2. `data/model_responses/title_mentions.csv`

### Evaluation

- [x] Notebook runs locally without error
- [x] Notebooks text is legible and sufficient
- [x] Notebook HTML snapshot looks reasonable


rlskoeser (approved):
Looks reasonable, let's merge.

The html notebook isn't interactive, which is fine for this. But maybe we can find some time soon to do experiment with [WebAssembly notebooks](https://docs.marimo.io/guides/wasm/#creating-wasm-notebooks).

PR #49, opened 2025-04-29
Pull over relevant setup files and data subsets from experiment 1

rlskoeser:
**Associated Issue(s):** NA

### Changes in this PR

- pre-commit hook configuration and setup documentation in the readme
- pyproject.toml and placeholder python package
- subset data files from previous round

### Notes

- I used git checkout to pull specific files from the ai-sandbox experiment branch that I thought would be useful

### Evaluation

- [x] Confirm you can install the empty package in this branch
- [x] Confirm pre-commit hooks are working correctly (I ran `pre-commit run --all`)


src/remarx/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.0.1.dev1"
```
laurejt, line 1: Do we want this to have the same version number as experiment 1?

laurejt (approved):

src/remarx/__init__.py
rlskoeser: Ah, thank you, I forgot! I remember we discussed differentiating those (although it likely doesn't matter much). Shall we call this `dev2` ?

laurejt: `0.0.1.dev2` sounds good to me

Issue #41, opened 2025-04-28
create a method combining lemmatization and regular expressions for identifying short phrases


mnaydan:
Use  [stanza](https://stanfordnlp.github.io/stanza/)

- identify candidate passages for title and concept passage mentions
- save results in csv or similar

Should help us find both positive and negative examples (esp for titles)

PR #51, opened 2025-05-01
Code for identifying candidate sentences containing key phrases using lemmatization

laurejt:
**Associated Issue(s):** #41 

#### Initial
- This includes initial code for finding sentences containing a passage of text that ignores inflection.
- Includes command-line script for finding title mentions for all texts within a directory and saves results as CSV
- Includes unused code for creating a sentence-based corpus
- Added resulting CSVs for Kapital and the Communist Manifesto for 1897-98aCLEAN, 1899-00a, and 1900-01a.

#### Update
- Refactored code for finding sentences containing a passage of text that ignores inflection
- Updated CSVs for Kapital and the Manifesto which now include a new column containing the list of found lemmatized phrases.

### Notes
- The script takes approximately 5 minutes per Die Neue Zeit text file.

### Evaluation
- [x] The output CSV structure and contents look reasonable
- [x] Can successfully run the script
- [x] Revised output CSV structure and contents look reasonable
- [x] Can successfully run the revised script

### DIFF
--- a/src/remarx/candidate_mentions.py
+++ b/src/remarx/candidate_mentions.py
@@ -1,0 +1,37 @@
+"""
+This module contains functions for identifying (candidate) sentences that
+might contain references to key titles or concepts.
+"""
+
+from typing import Generator
+
+import stanza
+
+NLP = stanza.Pipeline(lang="de", processors="tokenize,mwt,pos,lemma")
+
+
+def get_lemmatized_form(text: str, drop_punct: bool = True) -> str:
+    """
+    Converts a text into its lemmatized form. By default, punctuation is removed.
+    """
+    lemmas = []
+    for sentence in NLP(text).sentences:
+        for word in sentence.words:
+            if word.pos != "PUNCT" or not drop_punct:
+                lemmas.append(word.lemma)
+    return " ".join(lemmas)
+
+
+def get_sentences_with_key(key: str, text: str) -> Generator[str, None, None]:
+    """
+    Returns a generator of sentences within a given text that contain the
+    given search ignoring inflection.
+    """
+    # Lemmatize key
+    key_lemmas = get_lemmatized_form(key)
+
+    # Lemmatize & check each sentence
+    for sentence in NLP(text).sentences:
+        sent_lemmas = get_lemmatized_form(sentence.str)
+        if key_lemmas in sent_lemmas:
+            yield sentence.str
--- a/src/remarx/candidate_mentions.py
+++ b/src/remarx/candidate_mentions.py
@@ -90,7 +90,7 @@ def find_sentences_with_phrase(
     # Lemmatize & check if a sentence contains a search phrase
     for sentence in doc.sentences:
         lem_sent = lemmatize_sentence(sentence).lower()
-        if " " + lem_phrase + " " in " " + lem_sent + " ":
+        if f" {lem_phrase} " in f" {lem_sent} ":
             yield sentence
 
 
--- a/src/remarx/candidate_mentions.py
+++ b/src/remarx/candidate_mentions.py
@@ -24,7 +24,6 @@ import csv
 import pathlib
 import sys
 from collections.abc import Iterable
-from typing import Generator
 
 from stanza import DownloadMethod, Pipeline
 from stanza.models.common.doc import Document, Sentence
@@ -36,18 +35,25 @@ LANG = "de"
 PROCS = "tokenize,mwt,pos,lemma"
 
 
-def lemmatize_sentence(sentence: Sentence, drop_punct: bool = True) -> str:
+def lemmatize_sentence(
+    sentence: Sentence, drop_punct: bool = True, lowercase: bool = True
+) -> str:
     """
     Converts a stanza sentence into its lemmatized form with each word
-    separated by a space. By default, punctuation is removed.
+    separated by a space. By default, punctuation (as determined by part-of-speech)
+    is removed and resulting string is lowercased.
     """
-    return " ".join(
-        [w.lemma for w in sentence.words if w.pos != "PUNCT" or not drop_punct]
+    result = " ".join(
+        w.lemma for w in sentence.words if w.pos != "PUNCT" or not drop_punct
     )
+    return result.lower() if lowercase else result
 
 
 def lemmatize_text(
-    text: str | Document, pipeline: None | Pipeline = None, drop_punct: bool = True
+    text: str | Document,
+    pipeline: None | Pipeline = None,
+    drop_punct: bool = True,
+    lowercase: bool = True,
 ) -> str:
     """
     Lemmatizes a text using `lemmatize_sentence` with sentences separated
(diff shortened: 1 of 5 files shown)


rlskoeser:
>     * How do you think such results should be saved? At what level of aggregation?
>       
>       * This could easily operate over a collections of texts and/or multiple phrases.
>       * I've been thinking of the results operating at the key phrase + sentence level. So, a sentence with multiple references could appear multiple times.

It would be nice if we can generate a single csv output to aggregate results for however many terms we search for. With that goal in mind, output should include at least the following:
- search term
- filename
- sentence that includes the search term

What else would be useful? Would it be helpful to include character offsets for the sentence within the file, and can we easily generate them?

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    lemmas = []
+    for sentence in NLP(text).sentences:
+        for word in sentence.words:
+            if word.pos != "PUNCT" or not drop_punct:
+                lemmas.append(word.lemma)
+    return " ".join(lemmas)
+
+
+def get_sentences_with_key(key: str, text: str) -> Generator[str, None, None]:
```
rlskoeser, line 25: The word `key` here is throwing me off, I associate it with lookups like a dictionary. Phrase? Search term?

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,37 @@
...
+    given search ignoring inflection.
+    """
+    # Lemmatize key
+    key_lemmas = get_lemmatized_form(key)
+
+    # Lemmatize & check each sentence
+    for sentence in NLP(text).sentences:
+        sent_lemmas = get_lemmatized_form(sentence.str)
+        if key_lemmas in sent_lemmas:
+            yield sentence.str
```
rlskoeser, line 37: This could probably be a one-liner

rlskoeser (commented):
Looks like a good start.

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,37 @@
...
+"""
+
+from typing import Generator
+
+import stanza
+
+NLP = stanza.Pipeline(lang="de", processors="tokenize,mwt,pos,lemma")
+
+
+def get_lemmatized_form(text: str, drop_punct: bool = True) -> str:
```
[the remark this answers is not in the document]
laurejt: I mean... you can't really lemmatize sentences... But sure

src/remarx/candidate_mentions.py
laurejt: so you want search key?

src/remarx/candidate_mentions.py
laurejt: Perhaps, but that might make it completely illegible... and doesn't really make sense with yields

laurejt:
> It would be nice if we can generate a single csv output to aggregate results for however many terms we search for. With that goal in mind, output should include at least the following:
> - search term
> - filename
> - sentence that includes the search term

One issue is whether there can be multiple search terms for the same title / concept (which is yes); so there's a question of whether even more aggregation needs to occur.

> Would it be helpful to include character offsets for the sentence within the file, and can we easily generate them?

Possibly? We'd have to do the alignments ourselves. Also any text preprocessing could cause headaches.

rlskoeser:
> One issue is whether there can be multiple search terms for the same title / concept (which is yes); so there's a question of whether even more aggregation needs to occur.

Good point. So include both canonical term and then the variant form for the particular search?

laurejt:
> > One issue is whether there can be multiple search terms for the same title / concept (which is yes); so there's a question of whether even more aggregation needs to occur.
> 
> Good point. So include both canonical term and then the variant form for the particular search?

This seems like a later aggregation step, at least for the moment. Because the inputs for that would be a lot more complicated for the command-line script (maybe key-value pairs?). For now, I just operated at the title level.

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
+"""
+This script identifies candidate sentences within a collection of texts
+that contan phrases of interest (e.g., title or concept references)
```
rlskoeser, line 3:
```suggestion
that contain phrases of interest (e.g., title or concept references)
```

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+    # Lemmatize search phrase
+    lem_phrase = lemmatize_text(search_phrase, pipeline=pipeline).lower()
+
+    # Construct stanza document if needed
+    doc = pipeline(text) if isinstance(text, str) else text
+
+    # Lemmatize & check if a sentence contains a search phrase
+    for sentence in doc.sentences:
+        lem_sent = lemmatize_sentence(sentence).lower()
+        if " " + lem_phrase + " " in " " + lem_sent + " ":
```
rlskoeser, line 93: What are the spaces for? Is this to avoid partial word matches?

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+            doc = pipeline(text_file.read_text())
+            for i, sent in enumerate(doc.sentences):
+                # Lemmatize & lowercase sentence
+                lem_sent = lemmatize_sentence(sent).lower()
+                # Determine which phrases the sentence contains
+                matches = []
+                for j, phrase in enumerate(lem_phrases):
+                    # Add spacing to ensure word-level boundaries
+                    if f" {phrase} " in f" {lem_sent} ":
+                        matches.append(search_phrases[j])
```
rlskoeser, line 139: I see now from looking at the output that I wasn't understanding your comment about aggregation - of course sentences that match the more specific variant will also match the less specific case. If we knew the more specific variant you could privilege that one first and not bother checking others, but I like the format you came up with for the output (and the extra check is pretty trivial).

src/remarx/candidate_mentions.py
rlskoeser: Ah, I see similar logic below with a comment. The version with f-strings you have there is more readable.

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Construct stanza document if needed
+    doc = pipeline(text) if isinstance(text, str) else text
+
+    return " ".join(
+        [lemmatize_sentence(s, drop_punct=drop_punct) for s in doc.sentences]
+    )
+
+
+def find_sentences_with_phrase(
```
rlskoeser, line 70: It looks like this method is no longer used. I suggest removing it to avoid confusion

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+            file_progress.set_description(f"Processing {text_file.stem}")
+            doc = pipeline(text_file.read_text())
+            for i, sent in enumerate(doc.sentences):
+                # Lemmatize & lowercase sentence
+                lem_sent = lemmatize_sentence(sent).lower()
+                # Determine which phrases the sentence contains
+                matches = []
+                for j, phrase in enumerate(lem_phrases):
+                    # Add spacing to ensure word-level boundaries
+                    if f" {phrase} " in f" {lem_sent} ":
```
rlskoeser, line 138: It looks like you're assuming no punctuation for this method - totally fine by me. If we ever revisit and want that option, we should use a regex search with word boundaries.

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+
+    * file: name of text file
+    * sent_idx: sentence-level index within text file
+    * char_idx: character-level index within text file
+    * sentence: sentence text
+    * phrases: list of matching phrases
+    """
+    # Initialize stanza pipeline
+    pipeline = Pipeline(
+        lang=LANG, processors=PROCS, download_method=DownloadMethod.REUSE_RESOURCES
```
rlskoeser, line 116: This reuse download method is nice, will add to my code

src/remarx/candidate_mentions.py
```diff
@@ -0,0 +1,199 @@
...
+
+    # Lemmatize & lowercase pharses (for efficiency)
+    lem_phrases = [lemmatize_text(p, pipeline=pipeline).lower() for p in search_phrases]
+
+    fieldnames = ["file", "sent_idx", "char_idx", "sentence", "phrases"]
+    with open(output_csv, mode="w", encoding="utf-8-sig", newline="") as f:
+        writer = csv.DictWriter(f, fieldnames=fieldnames)
+        writer.writeheader()
+
+        file_progress = tqdm(input_texts)
```
rlskoeser, line 127: Would be nice if there was a way to show progress by sentence, or some indicator when the text has been pocessed. The tqdm output isn't granular enough for that. It doesn't look like the stanza pipeline has a way to any kind of streaming, though.

rlskoeser (approved):
I like the renamed methods, much clearer. 

Was able to run the candidate mentions script and generate output for a single input text. I added the file I generated.

What is the sentence corpus script for? Should I test running that also?

I think it's good enough to merge, but recommend removing the unused method first.

src/remarx/candidate_mentions.py
laurejt: I agree it would be nice, but to my understanding this isn't possible.

I can speak from recent experience, that pre-chunking the sentences only somewhat works (i.e., running the tokenizer-only and saving sentences to file) because these sentences might be further tokenized into multiple sentences when feeding them to stanza a second time *and* this runs at a much slower rate than doing it in the 5-minute chunks. That said, five minutes per-document is pretty good.

src/remarx/candidate_mentions.py
laurejt: I argue that we should keep it because this is a much more useful method if it's being invoked programmatically (e.g., in a notebook). Initially, I was using it for the command-line script but the extra calls to stanza were hurting performance.

laurejt:
> What is the sentence corpus script for? Should I test running that also?

This is residual code that seemed worth saving in case we want it later. The reason it's not currently in use is that to use the sentence corpus as input instead of the input directory appears to cause a performance hit because of the additional calls to the stanza pipeline (at least that's my best guess). Maybe it's worth a quick run, so we don't forget that it's untested.

> I think it's good enough to merge, but recommend removing the unused method first.

I argue that we should keep the method, it's much more useful for programmatic use, like the notebook I was initially using. It's not being invoked by the command-line script, because there's too much of a performance hit at scale.

src/remarx/candidate_mentions.py
laurejt: Thanks for catching this, I forgot I hadn't corrected it everywhere.

src/remarx/candidate_mentions.py
laurejt: ...and by lemmatization I mean part-of-speech

rlskoeser: Thanks for reminding, I saw that before but forgot it was punctuation based on part-of-speech tagging rather than the regex /character style of regex filtering.  

That makes me think it might be better to switch a regex word-boundary search sooner than later, though...

Issue #42, opened 2025-04-28
create notebook that runs out-of-the-box NER


mnaydan:
Run and compare stanza and flair; results of notebook should enable us to come to a decision about which one (if either) we want to use for other tasks.

- [stanza](https://stanfordnlp.github.io/stanza/)
- [flair](https://github.com/flairNLP/flair)

## todos
- [x] show a few examples for each
- [x] run on all title mentions subset
- [x] report on and compare matches
- [x] report on and compare sentences with no misc entities found
- [x] save results for flair as a baseline comparison for later work

rlskoeser:
completed in #52

PR #52, opened 2025-05-01
Notebook to try Stanza and Flair German NER for title mentions #42

rlskoeser:
**Associated Issue:** #42

### Changes in this PR

- Added stanza and flair to project dependencies
- New notebook trying out stanza and flair German NER
- **Updates since first review:**
  - Dropped unneeded lemmatization, added download reuse 
  - Added in-context highlighting to examples shown previously
  - Ran both NER models against all rows of the title mention subset, and output in tandem with the tags in that dataset for comparison about what titles we expect to find
  - Output the title mentions subset without annotation — not sure why some are found and not others 

### Notes

- I couldn't install flair on python 3.13 - seems to be a known issue with the sentencepiece dependency
- Let me know what more evaluation/context would be helpful.
  - ~Should I run on more examples now that I've got the basics working?~
  - ~Should I run it on all of the title mentions subset?~
  - Do we want to save results anywhere? 
  - Would it be helpful to add highlighting to show the entities in context?

### Evaluation

- [x] Confirm you can install dependencies and run the notebook
- [x] Confirm the html snapshot is present and accurately captures the content
- [x] Confirm explanatory text so far makes sense
- [x] Confirm you can run the updated notebook
- [x] Confirm revised output is more useful for assessing and comparing performance of Flair and Stanza on our title mention subset data
- [x] Confirm snapshot includes recent updates
- [x] Give feedback on what fields should be saved
- [x] Review data file capturing flair NER results for title mention subset


notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+    )
+    return
+
+
+@app.cell
+def _():
+    import stanza
+
+    nlp_de = stanza.Pipeline("de", processors="tokenize,mwt,lemma,ner")
```
laurejt, line 37: I believe the NER [Processor](https://stanfordnlp.github.io/stanza/pipeline.html#processors) only requires `tokenize` and `mwt`. If lemmatization is desired, you'll need to also include the `pos` processor.

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+            {"Text": ent.text, "Entity Type": ent.type} for ent in nlpdoc.ents
+        ]
+        ent_table = mo.ui.table(data=entity_data, pagination=True, selection=None)
+        return mo.vstack([text, ent_table])
+    return (show_stanza_entities,)
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""The first example sentence only identifies Marx and Engels as person entities, does not identify any titles.""")
```
laurejt, line 74: What titles are we hoping to find within the sentence?

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+
+@app.cell
+def _(doc1, sample1, show_stanza_entities):
+    show_stanza_entities(sample1, doc1)
+    return
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(r"""In the second example sentence, one of our titles is picked up, in spite of the fact that it includes a line break and our weird continuation character.""")
```
laurejt, line 86: How many titles are we hoping to find? Did stanza find all of the expected titles in this case?

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+    show_stanza_entities(sample2, nlp_de(sample2))
+    return
+
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(
+        r"""
+        Just for comparison, I used Google Translate on the first sentence and ran Stanza English NER. This is a different model with a much larger number of entities; it picks up Communist Manifesto as a "Work of Art."
+        """
```
laurejt, line 101: But it misses Kapital, right?

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+
+@app.cell(hide_code=True)
+def _(mo):
+    mo.md(
+        r"""
+        ## Flair
+
+        Run the same two sample text passages through Flair German NER model.
+
+        This one is picking up at least some of our titles under the MISC tag.
```
laurejt, line 129: Could you be more specific. How many titles were identified vs. the number annotated / expected?

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+        ]
+        ent_table = mo.ui.table(data=entity_data, pagination=True, selection=None)
+        return mo.vstack([text, ent_table])
+    return (show_flair_entities,)
+
+
+@app.cell
+def _(sample1, show_flair_entities):
+    show_flair_entities(sample1)
+    return
```
laurejt, line 185: This found both titles, right?

notebooks/try-ner.py
```diff
@@ -0,0 +1,195 @@
...
+
+
+@app.cell
+def _(sample2, show_flair_entities):
+    show_flair_entities(sample2)
+    return
```
laurejt, line 191: This found the one expected title, right?

notebooks/try-ner.py
laurejt, line 1: Looks good overall, but I think it's worth being explicit about what titles we want / expect to find vs. what the NER models identify.

rlskoeser:
@laurejt updated the notebook to run both stanza and flair on all rows of the title mention subset data and show rows with/without MISC annotations. I'm displaying with the tags you added, which indicates which titles we expect to be found.

We discussed saving the output from Flair as a baseline for comparison with later work. What fields should be included?
- uuid for title mention annotation
- text for items annotated as MISC entities
- NER model used? (e.g. "de-ner-large" for current tests)
- flair version? 

Should we duplicate relevant title mention fields (which might be all of them?) or assume the two files will be used together and we just need an id for joining?

laurejt:
> Should we duplicate relevant title mention fields (which might be all of them?) or assume the two files will be used together and we just need an id for joining?

I'd say for the purposes of this experimental phase, sentence-level IDs should be fine.

laurejt (approved):
Looks good overall, but make sure to resolve the `MapWithoutReturnDtypeWarning` warnings before merging.

rlskoeser:
@laurejt I've updated to fix the warnings and remove the redundant text (thanks for catching), and added a data file - I decided to output all entities found so we have a record — maybe it can help decide whether to include ORG and MISC tags both. (It's only a few cases so may not matter much.)

notebooks/try-ner.py
```diff
@@ -191,5 +324,103 @@ def _(sample2, show_flair_entities):
...
+
+
+@app.cell
+def _(flair_misc_ents, pl):
+    flair_misc_ents.select(
+        pl.col("Marx Title Mentions"),
+        pl.col("flair"),
+        pl.col("stanza"),
+    )
+    return
```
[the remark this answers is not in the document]
rlskoeser: The text column is displayed in the table just above — I omitted them because otherwise the table gets too wide.

laurejt:
> @laurejt I've updated to fix the warnings and remove the redundant text (thanks for catching), and added a data file - I decided to output all entities found so we have a record — maybe it can help decide whether to include ORG and MISC tags both. (It's only a few cases so may not matter much.)

I think saving all entities is the right call. We can disambiguate later.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ build-backend = "hatchling.build"
 [project]
 name = "remarx"
 description = "Code for identifying Marx citations associated with the CDH project Citing Marx"
-requires-python = ">=3.12"
+requires-python = ">=3.12,<3.13"
```
laurejt, line 8: Apparently, the 3.13 issue is actually a pip-specific problem. The conda-forge version of sentencepiece works on 3.13 (see [here](https://github.com/google/sentencepiece/issues/1083)).

Maybe another time to revisit how we're building our python environments?

laurejt:
> @laurejt you asked me yesterday if the titles ever showed up as any entities other than MISC and it occurred to me I hadn't tested that very well (only on a couple of examples). I ran a quick check and from the title mention subset I have a couple cases where Kapital and Die Neue Zeit are annotated as ORG. 😕
> 
> Any thoughts on how to reflect that in this notebook? It matters more for the next step we planned yesterday but feel like there should be some record of it here.

@rlskoeser I'd add another small section to the notebook walking through an example or two where this happens.

data/title_mentions_flair_ner.csv
laurejt, line 1: The CSV looks a bit funny to me. Empty fields are represented by `""`

notebooks/__marimo__/try-ner.html
laurejt, line 1: This is a stale snapshot.

laurejt (changes requested):
The snapshot is not up-to-date and the output looks a bit off since it's storing empty strings for some fields.

pyproject.toml
rlskoeser: I'll remove the version constraint

data/title_mentions_flair_ner.csv
rlskoeser: Apparently that is what Polars does for empty strings; I found and added a line to convert empty strings to `None` before saving.

laurejt (approved):

PR #54, opened 2025-05-02
Notebook for simple quote detection

laurejt:
**Associated Issue(s):** #53 

### Changes in this PR

- Contains notebook and snapshots documenting a simple method for detecting general quotations (via regex) 

### Notes

- I included both an HTML and Jupyter notebook snapshots. Let me know if the Jupyter notebook snapshot isn't useful.

### Evaluation

- [x] Can run the notebook locally
- [x] The notebook's contents are legible and sufficient
- [x] The HTML snapshot matches the notebook (at least from a static viewpoint)
- [x] The Jupyter snapshot matches the notebook. Comment on whether it is useful in addition to the HTML snapshot.


notebooks/find-quotes.py
```diff
@@ -0,0 +1,155 @@
...
+    # Highlight (and bold) Page Text
+    highlighted_page = page_text
+    for match in sorted(matches, reverse=True, key=lambda x: x.span()):
+        start, end = match.span()
+        highlighted_page = (
+            highlighted_page[:start]
+            + "<mark>**"
+            + highlighted_page[start:end]
+            + "</mark>**"
+            + highlighted_page[end:]
```
rlskoeser, line 145: I'm not sure why, but the mark + bold in dark mode results in white text with yellow highlighting, which is unreadable. Suggest using only one of these (or pulling in the custom css for highlighting).

```suggestion
            highlighted_page[:start]
            + "<mark>"
            + highlighted_page[start:end]
            + "</mark>"
            + highlighted_page[end:]
```

notebooks/find-quotes.py
```diff
@@ -0,0 +1,155 @@
...
+@app.cell
+def _(mo, page_slider):
+    mo.vstack(
+        [
+            mo.md("Select Slider:"),
+            page_slider,
+            mo.md(f"Selected Page: {page_slider.value}"),
+        ]
+    )
+    return
```
rlskoeser, line 99: For notebook usability, it's easiest to use when the slider is as close as possible to the content it controls. I moved the slider down one cell and hid code for both the slider and the text output when I was reviewing, which made it easier.

rlskoeser (approved):
Looks reasonable. 

I don't think the jupyter notebook adds much in this case.

notebooks/find-quotes.py
laurejt: I guess this is a case of light vs. dark mode; for me the bold makes the text more legible within the highlight.

laurejt: Maybe a case for needing to create an issue to make the custom css highlighting more useful in light-mode and more generally accessible (i.e. colorblind friendly)

notebooks/find-quotes.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+@app.cell
+def _():
+    # Store notes for pages (index by page id)
```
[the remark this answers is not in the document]
laurejt: I'll update the comment because "local" doesn't really explain the source / authorship of the notes.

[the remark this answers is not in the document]
rlskoeser: yes, good point! would be great to be explicit about the fact that these are _your_ notes

notebooks/find-quotes.py
laurejt: I really just want our own "<highlight>" tag 😂

Issue #38, opened 2025-04-28
prepare Marx text for text reuse methods


mnaydan:
find context for the example and make a page of source texts.

Information for the XML corpus found here: https://drive.google.com/drive/u/0/folders/1Y9ocfyqRJrlAGf_6lSc5Uu36dEnsnEe9

- [x] basic xml to text
- [x] get page-level text files from xml for specific pages (or all if easier); would be nice if we can run with a list of desired pages; might be nice to save as a jsonl corpus of pages
- [x] figure out which pages we want based on annotation citation data

if xslt becomes a pain, consider switching to neuxml and adapt code written in ppa-django for eebo-tcp

PR #55, opened 2025-05-02
Script to do baseline tei-to-text transform on MEGA xml content

rlskoeser:
**Associated Issue:** #38

### Changes in this PR

- Adds saxon to python dependencies, to support xsl transformation
- Simple script to use saxon to apply xslt to an xml file and save the result
- Adds tei-to-text and associated common functions xslt from the [stylesheets provided by the TEI consortium](https://github.com/TEIC/Stylesheets)

### Notes

- To run: download this XML file from drive [MEGA_A2_B005-00_ETX.xml](https://drive.google.com/file/d/151cShKSuICAGvCfik8vtokL8TTTLrJgl/view?usp=drive_link) and then run with the tei-to-text.xsl based on the example usage at the top of the file
- The output is pretty messy and seems to be adding a lot of line breaks etc. 
- We probably want to ask the folks who shared the XML files if they can share their transform logic (whether XSLT or otherwise). Would appreciate your input on how urgent this is.

### Evaluation

- [x] Confirm you can run the script to convert MEGA TEI to XML to plain text
- [x] Review the text generated and see if this messy version is good enough for preliminary/experimental use or if we need to refine.


laurejt:
One thing I wanted to note is that this version appears to make add line breaks for any xml tag, so any italicized text is on its own line. It's not the end of the world, but something that seemed worth noting.

Issue #56, opened 2025-05-05
Create filtered set of annotation data for evaluation of candidate title mentions


rlskoeser:
- [x] download csvs from recogito for our three texts (1897-98aCLEAN, 1899-00a, and 1900-01a)
- [x] aggregate, filter based on tags
- [x] calculate span indices (may be able pull from quote subset notebook logic in experiment 1)
- [x] add fields indicating whether a sentence includes each title (yes, no, maybe for Kapital and Manifesto; maybe ~ quotations)
- [x] save result to repo

PR #65, opened 2025-05-06
Add/update annotation data to experiment 2 branch

rlskoeser:
**Associated Issue:** ~ #56

### Changes in this PR

- Copied the annotation data from experiment 1
- Updated files with CSVs exported directly from recogito

### Notes

- Related to work & conversation on #56 

### Evaluation

- [ ] Quick review of list of files and expected fields


laurejt:
@rlskoeser Why do all these CSVs have empty strings `""`? This is not the case for the CSVs I downloaded from Recogito.

laurejt (approved):
This looks good, just note that some scripts operate on directories and so will include more than the three documents we're using for Title Mentions

rlskoeser:
> @rlskoeser Why do all these CSVs have empty strings `""`? This is not the case for the CSVs I downloaded from Recogito.

I don't know! I just used the CSV download button, I didn't modify them at all.

Issue #57, opened 2025-05-05
run flair on candidate title mentions data and save all the NER tags


rlskoeser:
Save as a new data file (CSV or JSON) in data repo for comparison with other methods. Store all the MISC spans - include the text as well as the start and end character ind.ces.

Preserve page id and sentence index for linking back to candidate mention data.

May be able to adapt existing method in the `try-ner` notebook.

laurejt:
Given the evidence that title mentions may appear as non-MISC spans, this will save all NER entity tags.

PR #68, opened 2025-05-06
Add script to generate ner spans & results CSV

laurejt:
**Associated Issue(s):** #57 

### Changes in this PR

- Script for saving NER spans from sentences file
- Resulting CSV for candidate_mentions/title_mentions.csv

### Notes

- This script saves all NER entities, not just MISC entities.
- The CSV only contains results for the three texts in the title mentions subset.
- Currently, the output CSV is encoded as `utf-8` if we thing this will be viewed in Excel we should consider encoding it as `utf-8-sig` (to add the Excel UTF-8 BOM).

### Evaluation

- [x] Check that this script can be run locally
- [x] Check that the csv output's fields and results look reasonable


src/remarx/get_ner_spans.py
```diff
@@ -0,0 +1,123 @@
+"""
+Script for gathering the NER spans of a collection of sentences (CSV) using flair.
+The input sentence file must have the following fields:
+file, sent_idx, char_idx, sentence.
+
+Examples:
+
+    python get_ner_spans.py canidate_sentences.csv candidate_ner.csv
```
rlskoeser, line 8:
```suggestion
    python get_ner_spans.py candidate_sentences.csv candidate_ner.csv
```

rlskoeser (approved):
Looks great. Runs well locally, output structure looks very usable.

Issue #58, opened 2025-05-05
join candidate mentions with annotation data


rlskoeser:
Create join utility (method) for matching candidate sentences with overlapping manual annotations.

Need a method to match annotations with candidate sentences based on span overlap. Try using a polars when join:
- start with the candidate sentences (expect multiple sentences may match to a single annotation)
- could end up with duplicates (if multiple overlap) - maybe combine the annotation tags and consolidate; order the title mentions (yes overrides a maybe)

Steps:
- [x] Create utility for loading candidate sentences and creating fields need for determining overlap and further evaluation
- [ ] Create method for joining candidate sentences with title annotations

PR #70, opened 2025-05-07
Utility for joining candidate sentences with title annotations

laurejt:
**Associated Issue(s):** #58 

### Changes in this PR

- Utility for loading title phrase mapping
- Utility for loading (and processing) of candidate sentences into a polars dataframe
- Utility for combining candidate sentences with annotations

### Notes

- The names for the module `remarx.polar_utils` and some of the methods may not be the best. Suggestions for alternative names are welcome.

### Evaluation

- [x] Check utility functions do the right thing when invoked within a notebook
- [x] Confirm the columns of the resulting dataframe of `annotate_candidate_sentences` are sufficient


src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
+import csv
+from collections import defaultdict
+
+import polars as pl
+
+
+def load_title_phrases(filename: str) -> dict[str, str]:
```
rlskoeser, line 7: The value of the returned dictionary is a list of strings, isn't it?

```suggestion
def load_title_phrases(filename: str) -> dict[str, list[str]]:
```

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+        for row in reader:
+            title_phrases[row["title"]].append(row["phrase"])
+    return title_phrases
+
+
+def load_candidate_sentences(
+    filename: str, title_phrases: dict[str, str]
+) -> pl.DataFrame:
+    """
+    Load from file and prepare candidate sentences dataframe.
```
rlskoeser, line 25: Suggesting a bit more detail for context of what goes in / what is expected out, based on my test run of this code in a notebook. Adapt as useful.

```suggestion
    Load candidate sentences from a CSV file and prepare a dataframe for combination with other data. Expects a CSV file generated by `candidate_mentions` script. Constructs a sentence id field and adds fields indicating which titles each sentence is a candidate for, based on the title phrase lookup passed in.
```

src/remarx/polar_utils.py
rlskoeser: Might be useful to set a default filename of `"data/title_searchphrases.csv"`, since we include that file in the repo and you're assuming those titles in the sentence preparation code below

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+            ).alias("kapital_candidate"),
+            ## Determine if sentence is a candidate for the Communist Manifesto
+            (
+                (
+                    pl.col("phrase_list").list.set_intersection(
+                        title_phrases["Manifesto"]
+                    )
+                ).list.len()
+                > 0
+            ).alias("manifest_candidate"),
```
rlskoeser, line 69: probably possible to do this as a conditional join (with a different structure for title phrases), but for the scale that we're working with for this experiment this is fine.

The logic is working correctly

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+        ]
+    )
+    return df
+
+
+# TODO: Rename method
+def annotate_candidate_sentences(
```
rlskoeser, line 76: here's an idea for a rename
```suggestion
def join_candidates_annotation(
```

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    candidate_sentences_df: pl.DataFrame, annotations_df: pl.DataFrame
+) -> pl.DataFrame:
+    """
+    Adds title mention annotations to candidate sentences dataframe. For a given
```
rlskoeser, line 80: probably worth noting: expects compiled annotations generated by `compile_title_annotations` script (correct?) and candidate sentences as loaded by the method above

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    # Sentences without annotations
+    unannotated_sents = candidate_sentences_df.filter(
+        ~pl.col("sent_id").is_in(annotated_sents.get_column("sent_id").to_list())
+    ).with_columns(
+        # add annotation-related columns
+        pl.lit("N/A").alias("uuid"),
+        pl.lit(-1, dtype=pl.datatypes.Int64).alias("anno_start_idx"),
+        pl.lit(-1, dtype=pl.datatypes.Int64).alias("anno_end_idx"),
+        pl.lit("No").alias("mentions_kapital"),
+        pl.lit("No").alias("mentions_manifest"),
```
rlskoeser, line 115: maybe anno prefix here also, to keep things explicit when looking at the combined dataframe

```suggestion
        pl.lit("No").alias("anno_mentions_kapital"),
        pl.lit("No").alias("anno_mentions_manifest"),
```

src/remarx/polar_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    # Sentences without annotations
+    unannotated_sents = candidate_sentences_df.filter(
+        ~pl.col("sent_id").is_in(annotated_sents.get_column("sent_id").to_list())
+    ).with_columns(
+        # add annotation-related columns
+        pl.lit("N/A").alias("uuid"),
+        pl.lit(-1, dtype=pl.datatypes.Int64).alias("anno_start_idx"),
+        pl.lit(-1, dtype=pl.datatypes.Int64).alias("anno_end_idx"),
```
rlskoeser, line 113: could these just be none / null? (probably doesn't matter for current experiment)

rlskoeser (approved):
Logic all looks fine here. I spun up a test notebook and tried these methods out - I've suggested documentation to make more explicit which files are expected to be used where.  It looks like the resulting combined annotations and candidates has all the information we'd need to compare and evaluate.

I have a few questions/suggestions, but I don't think anything needs to be changed (other than expanding the docs a bit, which is hopefully not too hard)

src/remarx/polar_utils.py
laurejt: It seemed like it'd be more of a pain if they were set to null for any operations we might want to do downstream... The set values make it clear it's not a valid span, but will also produce a reasonable length if we wanted to compile length calculations.

laurejt: Then again, this probably doesn't matter here. I'll switch the values to null for now.

Issue #69, opened 2025-05-07
Create notebook to compare sentence candidates with manual annotation.


jerielizabeth:
Want to identify cases where a title mention annotation is not in the candidate pool - i.e., lemmatization search failed
- ocr errors
- tagged as title mention but not directly cited or generic terms are used

Maybes = probably don't need to worry about for this phase

Issue #59, opened 2025-05-05
Compare flair results with known title mentions


rlskoeser:
Write a method to compare flair NER results from #57 with annotation data from #56 

The title mentions won't match exactly - need to figure out how to match them. We don't have spans for lemma search, but could make it easier if we refactor the search to add that — OR rerun the lemmatized search to find it at that point.

For matching entities with search matches: check the lemmas for span overlap; maybe a more simplistic lemma. This will probably need to be done in a notebook first to figure out what works.

Issue #44, opened 2025-04-28
create notebook for demo


mnaydan:

laurejt:
The CSV results are in the github repo, but have also been uploaded to the Project Google Drive Folder within Experiment II as `title_mentions_sent_results`

PR #72, opened 2025-05-20
Experiments/eval candidates

laurejt:
**Associated Issue(s):** #69, #59, #44

### Changes in this PR

- Notebook with evaluating our title mention detection methods (i.e., lemmatized search & NER) against manual annotations.
- CSV containing the collected results for each candidate sentence.
- Updated `compile title annotations` script to include the annotation's (span) text. Updated CSV accordingly.
- Added new utility methods to load title mention annotations and flair NER results as well as a method for joining the flair NER results to candidate sentence dataframes (can be annotated).

### Notes

- The notebook is incomplete. What seems most vital to add for the meeting?

### Evaluation

- [x] Check that the notebook successfully runs locally.
- [x] Check that the notebook snapshot matches the notebook.
- [x] Check that resulting `title_mentions_sent_results.csv` contents are legible and sufficient
- [x] Check that `compile_title_annotations` script runs locally and that its resulting CSV looks reasonable


src/remarx/polar_utils.py
```diff
@@ -32,100 +67,156 @@ def load_candidate_sentences(
...
+
+def join_candidates_ner(
+    candidate_sentences_df: pl.DataFrame,
+    flair_ner_df: pl.DataFrame,
+) -> pl.DataFrame:
+    """
+    Adds flair NER annotations to candidate sentences dataframe. Expects the flair
+    NER annotations to have been loaded using `load_flair_ner` and candidate sentences
+    to have been lodaded using `load_candidate_sentences` method. For a given
+    sentence without an annotation, teh following values are set:
```
rlskoeser, line 183:
```suggestion
    sentence without an annotation, the following values are set:
```

src/remarx/polar_utils.py
```diff
@@ -32,100 +67,156 @@ def load_candidate_sentences(
...
+        * flair_ner_spans: None
+        * flair_contains_band: false
+        * flair_contains_kapital: false
+        * flair_contains_manifest: false
```
rlskoeser, line 187: this documentation is really clear, thank you

rlskoeser (approved):
This all looks really good to me.

The `title_mentions_sent_results.csv`  data file has a lot of fields that probably needs filtering/explanation for other team members, but I understand it well enough and can't think of anything we need that isn't included.

Given that you're working on a different notebook for demo purposes, I think this is finished enough to merge.

Issue #35, opened 2025-04-28
run seq2seq-quotation-attribution and save the results


mnaydan:
https://github.com/uhh-lt/seq2seq-quotation-attribution

- [x] get setup to run on della
- [x] run on a few and review output
- [x] run on pages of text that have quotation annotations
- [x] minimal notebook with side by side annotation for review in context 
   - ~use different highlights for direct quote, indirect quote, and speaker~
   - [x] provide context information related to type of quote and other attributes

laurejt:
It looks like this code requires a specific computing setup not possible.

It might be possible to run this on `della` but pausing this effort for now.

rlskoeser:
I spent a little bit of time seeing if I could get this this installed on della for a test run.  I think I have it setup and ready to run, but need to write a slurm batch file to use a gpu and not ready to tackle that this late in the day. Here are some notes of what I've done so far:
- I cloned the entire repository (not included in their instructions but seems to be implied based on the other steps)
- I used mamba based on these instructions https://researchcomputing.princeton.edu/support/knowledge-base/python#mamba (not sure if I interpreted correctly, but I got it to work)
- Installed the dependencies as described in the docs
- Downloaded the linked model and expanded the tar file
- Downloaded the neue zeit text files from google drive and copied to della
- Set up the different directories as described (input text, temporary, output)

When I tried to run it with the partial-year issue level files, I got a warning about text file length:
> ValueError: [E088] Text of length 3012435 exceeds maximum of 1000000. The parser and NER models require roughly 1GB of temporary memory per 100,000 characters in the input. This means long texts may cause memory allocation errors. If you're not using the parser or NER, it's probably safe to increase the `nlp.max_length` limit. The limit is in number of characters, so you can check whether your inputs are too long by checking `len(text)`.

I downloaded one of the individual article files from elsewhere in drive and used that as my input file. I'm now able to run the predict command manually and can see that it is running. I get a CUDA warning (presumably because I'm not using a GPU), but it does start to process and I can see it's generating temporary output files.

I don't see any slurm files in the repo and don't know what resources it needs, but I can spend a little bit of time tomorrow running some test slurm jobs to get a sense of memory usage, etc, and see if I can get it to generate some results.

rlskoeser:
I successfully ran seq2seq on della for the single article file `Zinner_1896_LohnbewegungSchweizerischeEisenbahner.txt`

[...]

PR #73, opened 2025-05-21
seq2seq summary output, compilation notebook, review notebook, and improved highlighting styles

rlskoeser:
**Associated Issues:** 
- #35 seq2seq quotation output and review notebook
- #46 improve highlighting styles
- #45 demo notebook for quotation comparison

### Changes in this PR

- seq2seq quotes.csv file for results from running on 1896-97aCLEAN (chunked)
- input offset csv file for mapping seq2seq offsets for chunked files to full file, to match up with annotation data
- pulled highlight styles over from experiment 1 and updated to use color variables for two contrastive schemes and use  light-dark function
- seq2seq review notebook for working with the data, inspecting output, generating offsets, preliminary display
- quotation review/demo notebook for compilation to wasm
  - added symlinks to required data files in notebooks/public
  - requiring newer version of marimo for a bug fix to displaying polars dataframes in wasm notebooks

### Notes

- I haven't generated html snapshots for either of the notebooks yet. Is this still important?
- Here is the command I used for compiling a wasm version of the quotation-review notebook:
```console
marimo export html-wasm notebooks/quotation-review.py -o /tmp/marx-demo/ --mode run
```
You can put the output wherever you like. I always get a warning about ovewriting the index file, I think it must be a bug. After you run it, you'll see output directing how to serve it out with python locally
- I have manually copied the compiled wasm notebook to the pages branch
- Should I provide documentation in the review notebook for how I split the file into chunks? (currently documented on the github issue) recalculating offsets requires files not added to version control, but hopefully we don't need to redo
- I thought about adding the overlapping quotation highlighting but ran out of time

### Evaluation

- [ ] Review the highlight styles file: is there enough documentation to revise the color variables if needed?
- [x] Check that seq2seq-review notebook runs without errors; should not recreate the input offset file
- [x] Check that quotation-review notebook runs without errors
- [ ] Check the three highlight colors and styles used for the 3-up annotation comparison view in light and dark mode for contrast.
- [x] Check that you can compile quotation-review notebook to wasm
- [x] Check the public version of the compiled wasm notebook here: https://princeton-cdh.github.io/citing-marx/exp2/quotation-review/


notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+    # show this one first, since the display is keyed to seq2seq pages
+    if show_seq2seq.value:
+        panels.append(current_seq2seq_page())
+    if show_annotations.value:
+        panels.append(current_selected_annotation())
+    if show_heuristic.value:
+        panels.append(current_page_heuristic_quotes())
+
+    mo.hstack(panels)
+    return
```
laurejt, line 432: There's some weird displaying behavior when the width is set to full. With this setting and two "modes" toggeld on, there is a column-size gap between the displayed texts.

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+@app.cell
+def _(pl, quotes_source_text):
+    # adjust offsets to match annotated file
+    from intspan import intspan
+
+
+    # use intspan to parse and adjust by file offset and then serialize back to a span
+    def adjust_span_offset(span, adjustment):
```
laurejt, line 111: I think it's quite valuable to be clearer about what kind of span we're working with. In this case it's a numeric span (represented as a string). I really do find type annotations helpful when reviewing methods.

Also, a better name for this variable is `offset`.

notebooks/seq2seq-review.py
laurejt, line 1: I don't think that methods should be defined within the notebook, it's much easier to review code in a python source file.

I'm guessing this is tied to a WASM limitation?

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+    ).map_rows(
+        lambda row: adjust_span_offset(row[0], row[1]), return_dtype=pl.String
+    )
+
+    quotes_adjusted = quotes_source_text.with_columns(
+        quote_offsets_adjusted=pl.Series(adjusted_offsets),
+        quote_offset_start=pl.Series(adjusted_offsets).map_elements(
+            first_offset, return_dtype=pl.Int32
+        ),
+    )
```
laurejt, line 139: This just adds additional columns to `quotes_source_text` as opposed to adjusting existing columns, right?

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+
+    quotes_pages = quotes_pages.with_columns(
+        quote_offsets_page=pl.Series(page_offsets)
+    ).sort("quote_offset_start")  # sort by order in the file
+
+    mo.ui.table(quotes_pages, page_size=15, selection=None)
+    return (quotes_pages,)
+
+
+@app.cell
```
laurejt, line 187: I think there should be a display utilities module in the package.

data/seq2seq/1896-97aCLEAN_parts_output/quotes.csv
laurejt, line 1: Are these character spans pythonic (i.e., is the ending index included or excluded)?

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+            # text before the mark
+            text_parts.append(text[previous_end:start])
+            # text to be highlighted
+            text_parts.append(f"<mark>{text[start:end]}</mark>")
+            # set previuos end to the portion after this span
+            previous_end = end
+        # append any text after the last highlighted portion
+        text_parts.append(text[previous_end:])
+        return "".join(text_parts)
+    return (highlight_spans,)
```
laurejt, line 206: I think there will be some off-by-one errors due to the discrepancies between intspan and Pythonic ranges

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+
+
+@app.cell
+def _(highlight_spans, mo, quote_slider, quotes_pages):
+    def display_seq2seq_page(quote):
```
laurejt, line 211: It'd be nice to have some documentation of what this is being displayed

notebooks/seq2seq-review.py
```diff
@@ -0,0 +1,436 @@
...
+
+    # calculate page offsets, then turn into span
+    quote_page_spans = (
+        quote_annotations.with_columns(
+            start_index=pl.col("start_index").sub(pl.col("page_start")),
+            end_index=pl.col("end_index").sub(pl.col("page_start")),
+        )
+        .with_columns(
+            page_span=pl.concat_str(
+                [pl.col("start_index"), pl.col("end_index")], separator="-"
```
laurejt, line 255: If this is for intspan, this will cause an off-by-one error if the original indices correspond to a pythonic range.

notebooks/quotation-review.py
```diff
@@ -0,0 +1,415 @@
...
+    return
+
+
+@app.cell
+def _(mo, pl):
+    # Load data
+    #
+    # Load seq2seq quotes output, and previously calculated offset adjustment to annotated file.
+
+    # public locaiton for compilation to wasm
```
laurejt, line 52:
```suggestion
    # public location for compilation to wasm
```

notebooks/quotation-review.py
```diff
@@ -0,0 +1,415 @@
...
+app = marimo.App(
+    width="full",
+    app_title="Compare quotation detection methods",
+    css_file="highlight.css",
+)
+
+
+@app.cell
+def _():
+    import marimo as mo
```
laurejt, line 13: I find it nice to put all imports at the top rather than scattered throughout the notebook

notebooks/quotation-review.py
```diff
@@ -0,0 +1,415 @@
...
+        adjusted_span = intspan([i + adjustment for i in list(span)])
+        return str(adjusted_span)
+
+
+    # use intspan to parse and get first offset value
+    def first_offset(span):
+        return list(intspan(span))[0]
+
+
+    # df.select([pl.struct(["ham", "spam", "foo"]).apply(my_complicated_function)])
```
laurejt, line 88: Seems extraneous

notebooks/quotation-review.py
laurejt, line 1: A lot of this code is just copied and pasted from `seq2seq-review.py`, right? See my comments from in that notebook. Maybe it's a WASM problem, but it seems like a good case for wanting to separate reusable code from the notebooks.

laurejt, line 1: The 2nd and 3rd colors for light mode are too similar for my liking (a darker blue and a lighter blue). Try switching the 3rd to a different color.

laurejt (changes requested):
Generally looks good but I think there may be off-by-one errors for some of the text span ranges due to the discrepancies between `intspan` intervals and pythonic ranges.

For WASM is there any possibility of pointing to local code (i.e., a directory called `remarx`)?

Is it possible to have the notebooks have a toggle for light mode and dark mode? I think this is an important thing to figure out for our WASM results, since these are stuck in whichever mode it was compiled as. For example, your compiled version (and the public link) are in dark mode, while the one I compiled locally is in light mode.

PR #74, opened 2025-05-21
title mentions demo

laurejt:
**Associated Issue(s):** #44 

### Changes in this PR
_Include all key changes in this pull request_

- Adds demo notebook for title mentions task
- Include auxiliary table saved in a separate notebook

### Notes

- Would a static snapshot be useful to include in this case?

### Evaluation

- [x] Check that the notebook runs locally


rlskoeser (approved):
I initially got an error/warning about bad markdown syntax and an unclosed quote, but when I investigated I couldn't find one and then the error went away when I re-ran the cell.

Thought about compiling to wasm and publishing on pages, but don't think it's worth the effort to switch the data to a public path we can publish with the notebook.

I haven't found myself referring to the html exports of the notebooks, so I don't think it's needed.

Issue #84, opened 2025-05-28
Nearest Neighbor Search with embeddings / sentences


jerielizabeth:
Packages to try
- #97 
- #98
- #99 
- [ ] others if time (scikitlearn)

Evaluate requirements of package

Output is: 
- for sentences in Marx, K nearest matched sentences, and score

Evaluation - hopefully reusable across methods
- for a known sentence, how far down is the desired/expected sentence? 
- we care about top x (score directly may not be as meaningful / may not provide a useful cutoff); how does it change when we adjust the model or normalization or pipeline

rlskoeser:
We've completed and merged the nearest neighbor / semantic similarity methods we wanted to try during this experiment, so I'm going to close this now as well.

PR #94, opened 2025-05-30
Try out Chroma for finding similar sentences

rlskoeser:
**Associated Issue:** #84

### Changes in this PR

- Add `chromadb` to requirements
- New notebook trying out chroma using existing sentence corpora and sentence embeddings

### Notes

- I found some the [configuration documentation for Chroma](https://docs.trychroma.com/docs/collections/configure), which has information about the options for the embedding space; the default is squared L2 norm, should I try cosine similarity?
- Should resulting data file be added to version control?
- Should I add a bit more to the notebook to compare with the known specific citations in recogito?

### Evaluation

- [x] Check that you can run the notebook locally
~- [ ] Review evaluation code and output~
~- [ ] Check that the evaluation output has a reasonable set of fields~
~- [ ] Share your opinion: should we save a data file the full set of results for similar sentences? If so, for both or only one direction of querying?~
~- [ ] Confirm README documents all added files~

### Updated Evaluation (for @rlskoeser)
- [x] Check that you can run the notebook locally
- [x] Review evaluation code and output
- [x] Check that the evaluation output has a reasonable set of fields
- [x] Confirm README documents all added files


laurejt:
For these embeddings we can use inner product (`ip`) which I imagine will be the fastest option compute-wise. Since the embeddings are l2-normalized, inner product is equivalent to cosine similarity.

notebooks/try-chroma.py
```diff
@@ -0,0 +1,136 @@
...
+    mo.md(
+        r"""Load sample sentence corpora and embeddings, and load into the chroma collection."""
+    )
+    return
+
+
+@app.cell
+def _(np, pl):
+    dnz_sentences = np.load(
+        "data/sentence-embeddings/dnz-sample-sents.npy", allow_pickle=True
```
laurejt, line 47:
```suggestion
        "data/sentence-embeddings/dnz-sample-sents.npy"
```
I don't think you need to set this...

notebooks/try-chroma.py
```diff
@@ -0,0 +1,136 @@
...
+@app.cell
+def _(mega_sent_df, pl, result):
+    result_df = pl.DataFrame(
+        data={
+            "id": mega_sent_df["sent_id"].to_list(),
+            "input_text": mega_sent_df["text"].to_list(),
+            "result_text": result["documents"],
+            "distance": result["distances"],
+            "result_id": result["ids"],
+        }
```
laurejt, line 127: I think it'd be useful to have a field indicating the embeddings used since we expect to vary this (i.e., the model and the distance measurement). The filename is probably enough for now.

notebooks/try-chroma.py
```diff
@@ -0,0 +1,136 @@
...
+    ).explode("result_id", "result_text", "distance")
+
+    result_df.write_csv("mega_dnz_chroma.csv")
```
laurejt, line 130: Is there a way to indicate the level of precision for the distance values?

notebooks/try-chroma.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@app.cell
+def _(mega_sent_df, pl, result):
+    result_df = pl.DataFrame(
+        data={
+            "id": mega_sent_df["sent_id"].to_list(),
+            "input_text": mega_sent_df["text"].to_list(),
+            "result_text": result["documents"],
+            "distance": result["distances"],
```
laurejt, line 125: These scores seem about right, random embeddings will have a fairly high level of similarity, so we'll be looking at a very narrow band.

laurejt (commented):
- It'd be useful to try both directions for the search since it won't be symmetric. In other words, the following does not hold: if $\bar{a} is the closest vector to \bar{b} then \bar{b} is the closest vector to \bar{a}.
- It'd be quite useful to have some metadata on which DNZ sentences contain quotes and which Marx sentences are quoted. Then we could hopefully compute how many nearest neighbors $k$ we need to find the annotated results. Then again for this small example we should probably calculate it in full (rather than approximately) if we can.

rlskoeser:
@laurejt revisions all look great! And glad to know about polars implode (makes sense that it exists). I've resolved the merge conflicts with the experiment 3 branch and checked everything you listed, so I'll go ahead and merge.

Issue #87, opened 2025-05-28
Create method to extract embeddings for a specific word


jerielizabeth:
Have sentence - run through bert - get embedding for token (Kapital) - do something (cluster them)

spearman correlation

PR #96, opened 2025-05-30
Script to extract token embeddings for a specific term

laurejt:
**Associated Issue(s):** #87 

### Changes in this PR

- Updated dependencies
- Added new script for extracting specific term's token embeddings
- Added some unit tests for some of this script's underlying methods
- Added JSONL version of Experiment 2's `title_mentions_sent_results.csv` called `title-mentions-sents.jsonl`
- Added script results for the term "Kapital" for the sentences in `title-mentions-sents.jsonl`
- Updated README

### Notes

- The default MT5-XL model is quite large, so this script could take a fair amount of memory
- Note that currently this script only supports the family of [MT5](https://huggingface.co/docs/transformers/main/en/model_doc/mt5#overview) models. That said, the main difficulty with extensions is primarily tied to aligning a sentences word tokens with a model's subtokens.
- This may not work for any sentence within _Die Neue Zeit_ articles since there appear to be some rogue characters that we will need to handle in the future.

### Evaluation

- [x] Check that the script's code and documentation looks reasonable
- [x] Check that this script can run locally
- [x] Check that README includes all added files and sufficiently describes them


src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+
+    i_subtoken = 0
+    j_token = 0
+    while i_subtoken < len(subtokens) and j_token < len(spans):
+        sub_start = subtoken_starts[i_subtoken]
+        sub_end = (
+            subtoken_starts[i_subtoken + 1]
+            if i_subtoken < len(subtokens) - 1
+            else len(subtokens)
+        )
```
rlskoeser, line 99: Why not calculate span start/end when you do the initial iteration to get subtoken start indices?

Is `len(subtokens)` the right fallback here?

src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+    tokenizer: PreTrainedTokenizer,
+    sentence: str,
+    spans: list[tuple[int, int]],
+) -> list[list[int]]:
+    """
+    For a sentence, determine the correspondence between each given span and
+    the given T5 tokenizer's subtokens. In cases where span and subtoken boundaries
+    are misaligned, all subtokens that contain chracters within the span are included.
+    Returns the span's subtoken positions as a list of lists of integers.
+    """
```
rlskoeser, line 71: This description was hard to understand; after reading through the code and then reminding me the context (i.e., the `get_term_spans` method just above and our goal), I think I'm understanding better.

I think it would be clearer if you describe starting from the input; I'll suggest a starting point draft to work with or reject as you see fit 😄 :

> Starting with a sentence and a list of spans for terms or phrases of interest within that sentence, returns a list of subtoken indices that have any overlap for each span in the initial input.

There's a typo in characters (mention in case you preserve that part of the description when you revise).

You said "subtoken position" here — it's the index in the list of subtokens for this sentence, right?  

Am I understanding correctly that we need to return the index rather than the token because of the way the pipelines work?

rlskoeser: I think this code assumes spans are sequential - seems reasonable (at least for this phase) but probably worth  documenting?

src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+            start = subtoken_starts[i - 1] + len(subtoken)
+        subtoken_starts.append(start)
+
+    i_subtoken = 0
+    j_token = 0
+    while i_subtoken < len(subtokens) and j_token < len(spans):
+        sub_start = subtoken_starts[i_subtoken]
+        sub_end = (
+            subtoken_starts[i_subtoken + 1]
+            if i_subtoken < len(subtokens) - 1
```
rlskoeser, line 97: This looping logic is a bit hard to follow - I think it can be done in a single loop, esp. if you can assume spans are sorted (or sort them to guarantee it).  I think this would be more readable because then the code is operating more in terms of objects than indices into the different lists of subtokens and spans.

- iterate over subtokens, keeping track of current character index in the sentence (cumulative subtoken length)
- create a list variable to hold subtoken indices for the first/current span
- if current subtoken has any overlap with current span, add to list
- as soon as you are past the first span, add the list of found subtokens to the result and start a new list for the next span
- end when you hit the end of the last span

src/remarx/get_token_embeddings.py
rlskoeser: Does method also assume spans don't overlap?

Is it ever possible for two subtokens to belong to two different spans?

rlskoeser (commented):
I've only made it part way through but commenting so you can see my notes and questions so far.

I'll look at `extract_span_embeddings` and `save_term_embeddings` next.

src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+def normalize_text(text: str) -> str:
+    """
+    Normalizes text so that it will align with its tokenized version.
+    """
+    # Remove trailing whitespace
+    result_text = text.rstrip()
+    # Replace consecutive whitespace with a space
+    result_text = " ".join(result_text.split())
+    # Add leading space
+    result_text = " " + result_text
```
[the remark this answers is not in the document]
laurejt: I think we can assume there isn't a leading space, but I realize I should just `strip`. I currently have the logic separated out because we may need more steps (I didn't look that closely at T5's preprocessing). Also, I find it more legible to read the steps separately rather than a single one-liner return.

src/remarx/get_token_embeddings.py
laurejt: This isn't really a computation. It's a lookup.

No, it's not a fall back; it's a typo. It should be `len(sentence)`. The end character index for each subtoken is the start of the next subtoken except the final token whose end is the length of the string (since there is no next subtoken)

src/remarx/get_token_embeddings.py
laurejt: There are some assumptions, but spans will be sorted (in ascending order) and non-overlapping by how `re.finditer` works.

src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+def get_term_spans(text: str, term: str) -> list[tuple[int, int]]:
+    """
+    Finds all instances of the provided term within the input text.
+    Returns a list of span start and end indices for each identified instance.
+    """
+    pattern = r"(\b|^)" + term + r"(\b|$)"
+    term_indices = []
+    for match in re.finditer(pattern, text):
+        term_indices.append(match.span())
+    return term_indices
```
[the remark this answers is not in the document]
rlskoeser: ah, ok! I glanced at the docs but didn't see that difference clearly enough, makes sense now

src/remarx/get_token_embeddings.py
```diff
@@ -0,0 +1,301 @@
...
+    """
+    Normalizes text so that it will align with its tokenized version.
+    """
+    # Remove trailing whitespace
+    result_text = text.rstrip()
+    # Replace consecutive whitespace with a space
+    result_text = " ".join(result_text.split())
+    # Add leading space
+    result_text = " " + result_text
+    return result_text
```
rlskoeser, line 46: one-line version, fwiw. I hadn't thought about adding more steps later, but for this set of steps it seems pretty readable.

```suggestion
    # Remove any outer whitespace, consolidate multiple whitespace characters, and add a leading space
     return " " + " ".join(text.strip().split())
```

laurejt: You've convinced me. Thanks.

src/remarx/get_token_embeddings.py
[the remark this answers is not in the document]
laurejt: I too thought I should be using `findall` before I realized it just gave me strings.

src/remarx/get_token_embeddings.py
laurejt: Yes, subtoken position corresponds to index within the list; wasn't sure what the best / most concise way would be.

rlskoeser: These are reasonable assumptions, esp. as it's only used by the calling code and we're constructing the spans! I just wanted to make sure I understood the assumptions. It seems like we should be able to avoid generating overlapping spans.

Based on your knowledge of how tokenization works, is it the case that we shouldn't expect a subtoken to overlap with two different spans we care about?

src/remarx/get_token_embeddings.py
rlskoeser: I don't mean to get us off track with optimizing / refactoring we don't need for an experiment, just want to understand what is going on - if we're confident it is working as needed then there is no need to revise now! And I do think I understand what's going on well enough. But if we end up having to revise it for the experiment, refactoring might help.

src/remarx/get_token_embeddings.py
laurejt: I don't believe so but we'll have to keep the strong assumptions on which pre-trained tokenizer we used; many tokenizers tend to split on whitespace (and generally on punctuation but that's less of a guarantee). I can say that some don't but I don't believe we'll encounter those.

I'm taking a closer look at the pretrained T5 tokenizer to more fully understand the pretokenization process used by the tokenizer as well as some possibilities for preserving our tokenization. Granted we'd actively have to predefine that word-level tokenization process and see the potential consequences (all of which require additional work)

laurejt: From what I can tell, I don't think we'd have a subtoken contribute to two spans we care about. While I don't think there's anything inherently preventing sentencepiece/unigram subtokens from containing internal whitespace, I'm guessing this wouldn't occur through shere probability.

laurejt: Note that while I don't think that multiple spans will match to the same subtoken, the code should allow for this case.

rlskoeser (approved):
Was able to run the script locally on a few different sentence corpus files and with a few different terms. Was initially surprised that the token index in the CSV file was zero for all instances of Kapital, but figured out that this is the index for tokens of interest, e.g. if the term of interest occurs more than once in a single sentence.

Hit a few small snags running it (e.g. needing to create the output directory first), but the script error messages i ran into were sufficient to figure everything out.

PR #102, opened 2025-06-03
Sentence evaluation data

rlskoeser:
**Associated Issue:** #84

### Changes in this PR

- Revisions to the tei text extraction script to support identifying cited passages:
  - add option to output line numbers
  - add option to omit footnote contents (references and note text)
  - make output file optional; if not specified, outputs text to the console 
- New MEGA sample text file `data/mega-sample-pages/p395-396.txt` - pages 395 and 396, generated without footnotes, since one of our example citations spans these pages
- Preliminary notebook to convert annotation data into evaluation data:
  -  load annotation data and match with DNZ sentences based on offsets
  - output quotation text and comment with citation details in a format to support manually collecting cited text
  - manually collected text based on the citation page and line numbers and the revised tei text script. 
  - todo list of what I think are the remaining steps needed to turn this into DNZ <> MEGA sentence pairs we can use for evaluation
 
**Updates**:
- Removed the p396.txt data file 
- Regenerated MEGA sentence corpus file and sentence embeddings from revised text content
- Added sentence pair candidate files and preliminary set of matching sentences

### Evaluation

- [x] Review tei script changes and test locally that it works with the new options (see below)
- [x] Check that the notebook runs locally
- [x] Check offset overlap logic in notebook for matching annotations to DNZ sentences
- [x] Review manually collected lines of MEGA 
- [x] Review revised notebook for identifying dnz and mega sentences
- [x] Review data files and confirm paired sentences file has needed information
- [x] Confirm readme update includes all files included in this PR

---
Example TEI script  usage (adjust path as needed for your copy of MEGA TEI):
```console
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 395 -e 397 --no-footnotes -o data/mega-sample-pages/p395-396.txt
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 395 -e 397  -l --no-footnotes
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 402 -l --no-footnotes
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 25 -l --no-footnotes
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 25 -l
./src/remarx/tei_page.py data/MEGA_A2_B005-00_ETX.xml -s 147 -l
```


laurejt:
There is no modification / preprocessing of the DNZ texts, so the character indices should be exact.

laurejt (commented):
Given the added changes to parsing the Marx texts, it seems to me like we actually want a more structured data style approach. Perhaps we could save this as a CSV or JSONL? At a minimum, I see the following fields: page, line number, and whether the line is from the body of the text or a footnote. This could let us stitch the body of the text back together and let us preserve footnotes. Granted this approach would require a more custom generation of sentences, but that's seeming important for the future (but perhaps less relevant for this experiment).

laurejt:
Given that one of the quoted passages of interest spans across a page break, we clearly need to take footnotes into account. In the long run, I think we need to preserve the footnotes but perhaps we can separate them from the main body of the text. That said, it seems less valuable for the experiment.

notebooks/quotation-sentence-ids.py
```diff
@@ -0,0 +1,216 @@
...
+def _(dnz_sent_df, pl, quote_df):
+    # join subset of quotes with page text; rename page text columns for clarity
+    quote_sentences = quote_df.join_where(
+        dnz_sent_df,
+        # limit to annotations and sentences from the same file
+        pl.col("FILE") == pl.col("file"),
+        # look for sentences with any overlap with the annotation content
+        pl.col("start_index") <= pl.col("char_end_idx"),
+        # annotation ends after sentence starts
+        pl.col("end_index") >= pl.col("char_idx"),
```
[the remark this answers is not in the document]
rlskoeser: wow, thank you again!

notebooks/quotation-sentence-ids.py
```diff
@@ -132,17 +145,47 @@ def _(mo, quote_df):
...
+            "start_str": "Ist der Werth dieser Kraft höher",
+            "end_str": "in verhältnißmäßig höheren Werthe",
+        },
+    }
+
     # text content here was manually extracted using the tei_page script with line numbers turned on and footnotes omitted,
     # then manually selecting from the specified lines based on the quotation text above
 
-    mega_text = [
+    # this is the full set of text content used to generat start and end strings above;
```
laurejt, line 182:
```suggestion
    # this is the full set of text content used to generate start and end strings above;
```

notebooks/quotation-sentence-ids.py
```diff
@@ -166,33 +209,101 @@ def _(mo, quote_df):
...
+    cited_text_info = []
+
+
+    for annotation_id, citation_source in mega_text.items():
+        file_path = mega_page_dir / citation_source["file"]
+        file_contents = file_path.read_text()
+        start_index = file_contents.index(citation_source["start_str"])
+        end_index = file_contents.index(citation_source["end_str"]) + len(
+            citation_source["end_str"]
+        )
```
laurejt, line 256: Just noting that this could fail / produce erroneous indices if there is more than one occurrence of one of the strings

laurejt (approved):
Update the README to reflect the sentence pairs CSV and then 🚀

notebooks/quotation-sentence-ids.py
rlskoeser: yes - since it's such a small set, I worked on and checked them manually to get long enough strings to match the content I wanted to find; this approach wouldn't scale, but I don't expecting we'd repeat this.

Issue #83, opened 2025-05-28
Create methods to post-process embeddings


jerielizabeth:
- [x] dimensionality reduction
- [x] spearman-rank correlation

(scikit learn)

everything that does KNN will do cosine similarity and euclidean distance for us

once we have an evalation mechanism for top-n results then we can compare different post-processing steps

PR #103, opened 2025-06-12
Added several embedding transformation methods

laurejt:
**Associated Issue(s):** #83 

### Changes in this PR
- Add library with several embedding transformation methods
- Updated pyproject.toml with new dependencies

### Notes

- This is a collection of methods rather than a script. So, to check the validity of the methods, you'll need to run them in a separate script/notebook/python interactive shell.
- Are there any other transformations / post-processing methods that we need?

### Evaluation

- [x] Check that methods can be called from a notebook and/or python interactive shell. Check that the resulting arrays have the correct shape.
- [x] Check that the methods are sufficiently described

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,11 +27,15 @@ dependencies = [
   "intspan",
   "lxml",
   "sentence-transformers",
+  "scikit-learn",
+  "scipy-stubs", # for mypy
 ]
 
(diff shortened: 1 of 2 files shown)


rlskoeser (approved):
Was able to install and run all the methods in a notebook to test. Output looks good, docstrings seem sufficient.

Issue #99, opened 2025-06-02
Use annoy library for approximate nearest neighbors


laurejt:
https://github.com/spotify/annoy

- [ ] update readme
- [ ] add html snapshot

PR #109, opened 2025-06-26
Running & Evaluating Annoy

laurejt:
**Associated Issue(s):** #99 

### Changes in this PR
- Added annoy notebook for creating and evaluating results
- Added annoy results CSV
- Added annoy dependency to pyproject.toml

### Notes

- I've had some trouble with marimo reverting Markdown cells to code cells. If you find any, please try making the changes.

### Evaluation

- [ ] Check that the you can run the notebook locally
- [x] Check that the sentence-pair results contain the necessary fields
- [x] Check that the notebook's contents are legible and sufficient


notebooks/try-annoy.py
```diff
@@ -0,0 +1,401 @@
...
+    # Build knn results
+    entries = []
+    for dnz_i, dnz_vec in enumerate(dnz_embeddings):
+        knn_results = marx_index.get_nns_by_vector(
+            dnz_vec, n=10, include_distances=True
+        )
+        for rank in range(10):
+            # Marx sentence row index
+            marx_i = knn_results[0][rank]
+
```
rlskoeser, line 130: I'm getting an `IndexError` on line 129. I tried inspecting and it seems that I'm only getting one result per query, rather than the expected 10. Any ideas?

Issue #116, opened 2025-08-04
set up precommit hooks


mnaydan:
- Prettier
 - Codespell
 - ruff (use ruff configuration from pyproject.toml in #115 )

Note:

Prettier now has been replaced with:
- yamlfmt → YAML formatting 
- mdformat → Markdown formatting

Currently, we do not use a linter for TOML formatting because it significantly alters the file’s format from the team’s preferred format.

PR #151, opened 2025-08-05
Setup pre commit hooks

tanhaow:
Associated Issue(s): #116 

## Changes in this PR

- Adds `.pre-commit-config.yaml` with ruff, prettier, and codespell hooks
- Minor changes in `README.md`, `__init__.py`, and `pyproject.toml` with formatting from pre-commit hooks

## Notes
- Ruff is configured to run with `--fix` flag for automatic linting fixes
- Ruff is configured to run with configuration from `pyproject.toml`

## Reviewer Checklist

**@laurejt:**
- [ ] Check that pre-commit can be installed and initialized locally (`pre-commit install`)
- [ ] Check that pre-commit hooks run successfully (`pre-commit run --all-files`) 

**@rlskoeser:**
- [ ] Check that ruff rules are not too restrictive for development workflow


.pre-commit-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - id: ruff-format
+        types_or: [python, pyi, jupyter]
+
+  # Prettier for formatting various file types
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v4.0.0-alpha.8
+    hooks:
+      - id: prettier
+        types_or: [markdown, yaml, json, toml]
+        exclude: ^(\.git/|\.github/workflows/)
```
rlskoeser, line 24: I would prefer not to exclude github workflow files, sometimes we check in yaml files with formatting problems that mess up our github actions, would be great to catch at commit time instead.

rlskoeser (commented):
Was able to install pre-commit hooks and run them on everything. As a quick test, I modified the spacing and indentation in the pyproject.toml file expecting prettier to fix it - but it didn't. Do we need additional plugins?

.pre-commit-config.yaml
tanhaow: Good catch— it seems like Prettier [doesn’t format TOML](https://prettier.io/docs/) on its own, so the hook skipped pyproject.toml. I will find and add an additional plugin for TOML

laurejt:
Note for future branches. This should be a feature branch created in git flow.

README.md
```diff
@@ -21,14 +21,18 @@ This repo uses [git-flow](https://github.com/nvie/gitflow) branching conventions
   including all python dependencies and optional dependencies for development and testing:
 
 ```sh
+# Using uv (recommended)
+uv sync --extra dev
```
laurejt, line 25:
```suggestion
uv sync
```
Since we're using default groups, `uv` includes dev dependencies by default.

tanhaow: got it! Thanks!

Issue #117, opened 2025-08-04
As a user, I want a method to segment text into sentences so I can transform input into the required format for the core pipeline.


mnaydan:
## Testing Notes (QA):

Review the generated sentence corpus output from testing #122 (plaintext) and #131 (TEI file): 

(Note: Non-paragraph parts of the file might be messy.)

- [x] Find a section in the output for a regular paragraph, and check if each row in the spreadsheet is an individual sentence. 
- [x] Check if the sentences are sequential and match the content in the input paragraph.
- [x] Confirm that the text is not modified, but just chunked. 

---
- adapt from existing code in experiment 2
- add unit tests & docstrings
- most likely using stanza

Note: currently waiting for acceptance test

rlskoeser:
Thank you @bn2409-svg for the detailed testing notes, these are so helpful.

Several of the things you've flagged look related to the way we're converting the TEI to plain text, and this version is only the first pass which we've kept deliberately simple; we have work in progress now to split out footnotes from main body text, which should help with some of these things. I'd be a little surprised if italicized text is introducing line breaks in the plain text we're feeding into the sentence segmentation, although it's possible; it looks like the fraction is not being rendered properly. It would be good to create an issue to track these so we can keep an eye on them as potential enhancements; I'm not sure how crucial it is for our immediate goals. 

The way we've implemented things right now, you can't view the plain text generated from the TEI because we feed it directly into the sentence segmentation process. If reviewing the TEI to text conversion would be helpful, we can think about a simple way to give you access to that.

If you want to test how the much impact the line breaks have, you could construct a small-scale test with a plain text file that varies this (e.g., one version of the passage with with the semicolon at the end of the line, one version with the same text but no line breaks).

mnaydan:
As discussed with @bn2409-svg on Slack, we are going to close this issue and track enhancements on #197, including any additional things Ed notices once he gets a chance to look at the output. But based on Ben's testing, this is good enough to close for the 0.1 release.

PR #154, opened 2025-08-06
Feature/sentence segmentation

tanhaow:
**Associated Issue(s):** #117 

### Changes in this PR

- add sentence segmentation method and unit tests
- Added `tqdm` (progress bar) to `pyproject.toml`

### Reviewer Checklist

@laurejt 
- [x] Confirm the input and output of the functions are what we want

@rlskoeser 
- [x] Review code quality and architecture decisions
- [ ] Review unit tests --> Question: Should we add tests that use real Stanza models? (Currently all tests just use mocks to test the code logic and assume Stanza segments sentences well)

### DIFF
--- a/tests/test_sentence/test_segment.py
+++ b/tests/test_sentence/test_segment.py
@@ -75,8 +75,3 @@ class TestSegmentTextIntoSentences:
         segment_text("Hallo Welt.")
         mock_pipeline_class.assert_called_with(lang="de", processors="tokenize")
 
-    @patch("remarx.sentence.segment.stanza", None)
-    def test_segment_text_no_stanza(self) -> None:
-        """Test that function fails gracefully when stanza is not available."""
-        with pytest.raises(AttributeError):
-            segment_text("Test text.")


tanhaow:
Thanks for your code review, @laurejt . I’ll refactor the code so the function accepts only a `str` and returns the list of `(char_idx, sentence)` pairs. 
I originally included file handling logic because Experiment 2 included filename in its return value, but I see that’s outside the scope here. I’ll simplify the implementation and push the update soon.

rlskoeser:
> Thanks for your code review, @laurejt . I’ll refactor the code so the function accepts only a `str` and returns the list of `(char_idx, sentence)` pairs. I originally included file handling logic because Experiment 2 included filename in its return value, but I see that’s outside the scope here. I’ll simplify the implementation and push the update soon.

@tanhaow yes, sorry we weren't clearer - we did say adapt experiment 2 code and we didn't specify inputs and outputs in the issue or in conversation. I think @laurejt and I had it in our heads because we've been working on the technical design and thinking about other aspects of the code and the various input files we'll need to handle, but we hadn't articulated how it impacted this task.

We'd also like to restructure the remarx code into submodules, so please move your refactored code into `remarx/sentence/segment.py` and then move the test code to match.

tanhaow:
> > Thanks for your code review, @laurejt . I’ll refactor the code so the function accepts only a `str` and returns the list of `(char_idx, sentence)` pairs. I originally included file handling logic because Experiment 2 included filename in its return value, but I see that’s outside the scope here. I’ll simplify the implementation and push the update soon.
> 
> @tanhaow yes, sorry we weren't clearer - we did say adapt experiment 2 code and we didn't specify inputs and outputs in the issue or in conversation. I think @laurejt and I had it in our heads because we've been working on the technical design and thinking about other aspects of the code and the various input files we'll need to handle, but we hadn't articulated how it impacted this task.
> 
> We'd also like to restructure the remarx code into submodules, so please move your refactored code into `remarx/sentence/segment.py` and then move the test code to match.

Thanks for the clarification! With the inputs and outputs now specified, it's much clearer. I’ll simplify the function to process plain text only and move it to the appropriate place.

test/test_sentence_segmentation.py
```diff
@@ -0,0 +1,58 @@
+"""
+Unit tests for sentence segmentation functionality.
+"""
+
+from unittest.mock import Mock, patch
```
laurejt, line 5: Use `NonCallableMocks` instead for things that aren't methods (i.e., should never be called directly).

test/test_sentence_segmentation.py
```diff
@@ -0,0 +1,58 @@
...
+def create_mock_sentence(text: str, start_char: int = 0) -> Mock:
+    """Helper function: create a mock sentence with the given text and start character."""
+    mock_sentence = Mock()
+    mock_sentence.text = text
+    mock_sentence.tokens = [Mock()]
+    mock_sentence.tokens[0].start_char = start_char
+    return mock_sentence
+
+
+class TestSegmentTextIntoSentences:
```
laurejt, line 19: This should also be testing the `language` argument. Don't worry about unsupported languages, but it should be able to work whether "de" is explicitly provided or not.

laurejt:
@tanhaow The unit tests shouldn't run stanza, but it's worth testing when stanza is invoked (just patch as mocks, we did this for [corppa's ocr script](https://github.com/Princeton-CDH/corppa/blob/develop/test/test_ocr/test_gvision_ocr.py).

rlskoeser (changes requested):
Refactored method looks good!  Requesting some minor changes.

I think it makes sense not to test with stanza methods directly, since I don't know of a way to do that without downloading models, which we definitely don't want to do on GitHub Actions. But let's do take advantage of the `mock` capacity to use a class as a specification, so we're testing against the stanza code interface.

test/test_sentence_segmentation.py
```diff
@@ -0,0 +1,58 @@
+"""
```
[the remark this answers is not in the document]
tanhaow: but just noticed that [corppa](https://github.com/Princeton-CDH/corppa/blob/develop/test/test_ocr/test_gvision_ocr.py) actually used `test` as directory name :)

laurejt:
Remember to create / update the change log.

tests/test_sentence/test_segment.py
```diff
@@ -0,0 +1,82 @@
...
+
+        # Test with default language (should be "de")
+        segment_text("Hallo Welt.")
+        mock_pipeline_class.assert_called_with(lang="de", processors="tokenize")
+
+    @patch("remarx.sentence.segment.stanza", None)
+    def test_segment_text_no_stanza(self) -> None:
+        """Test that function fails gracefully when stanza is not available."""
+        with pytest.raises(AttributeError):
+            segment_text("Test text.")
```
rlskoeser, line 82: Good thought trying to think of other possible error cases. I don't think this could happen the way the code is currently written - the method won't be defined if the import fails, and we don't intend stanza to be an optional dependency.  
```suggestion
```

Issue #168, opened 2025-08-11
Improve mkdocs documentation with plugins


tanhaow:
- Create a context table for submodule API descriptions in the code documentation.
- Fix the relative path reference issue on the GitHub repository and the website.

PR #172, opened 2025-08-18
Enhance mkdocs

tanhaow:
**Associated Issue(s):** #168 

### Changes in this PR
This PR has two enhancements on remarx's documentation.

- The linking issue has been resolved. Initially, the link could either work on GitHub or the Mkdocs website. Now, it works on both. The issue was resolved by replacing the old symlink solution (docs/index.md → README.md) with a MkDocs hook that automatically changes links during the build process. Consequently, the GitHub README uses a link that links to the Developer Notes at the root directory, while the Mkdocs website README uses a link that links to the Developer Notes under the /docs directory to ensure the site renders correctly.

- An index table has been added to display the modules that Remarx has and their corresponding docstrings. This table was generated by the mkdocs-gen-files plugin and newly added MkDocs hooks.
   image
   
   
### Reviewer Check List
- [ ] run `uv run mkdocs build` to check if the site can be successfully built.
- [ ] run `uv run mkdocs serve` check if the site's layout and content seem reasonable.

### DIFF
--- a/src/remarx/sentence/corpus/tei_input.py
+++ b/src/remarx/sentence/corpus/tei_input.py
@@ -145,8 +145,9 @@ class TEIinput(FileInput):
 
     def __post_init__(self) -> None:
         """
-        After default initialization, parse the input file as a TEIDocument
-        and store it as xml_doc.
+        After default initialization, parse the input file as a
+         [TEIDocument][remarx.sentence.corpus.tei_input.TEIDocument]
+        and store it as [xml_doc][remarx.sentence.corpus.tei_input.TEIinput.xml_doc].
         """
         # parse the input file as xml and save the result
         self.xml_doc = TEIDocument.init_from_file(self.input_file)
--- a/src/remarx/sentence/corpus/__init__.py
+++ b/src/remarx/sentence/corpus/__init__.py
@@ -3,7 +3,7 @@ Functionality for loading and chunking input files for sentence corpus creation.
 """
 
 from remarx.sentence.corpus.base_input import FileInput
-from remarx.sentence.corpus.tei_input import TEIinput
+from remarx.sentence.corpus.tei_input import TEI_TAG, TEIDocument, TEIinput, TEIPage
 from remarx.sentence.corpus.text_input import TextInput
 
-__all__ = ["FileInput", "TEIinput", "TextInput"]
+__all__ = ["TEI_TAG", "FileInput", "TEIDocument", "TEIPage", "TEIinput", "TextInput"]


rlskoeser (commented):
The result is nice, but a little concerned about the amount of custom code needed to make it happen. Is this the recommended solution? Are there no existing plugins that do what we need?

If `docs/api/index.md` is auto-generated, it should not be checked into version control.  Similar comment about the devnotes file in the docs directory — can this be removed now?

laurejt:
@tanhaow I'll review once you create a check list.

DEVELOPERNOTES.md
laurejt, line 1: Why is this now an empty file?

laurejt:
Have the instructions change for previewing documentation locally? When I run `uv run mkdocs serve`, I get a 404 error for the developer notes.

tanhaow:
> Have the instructions change for previewing documentation locally? When I run `uv run mkdocs serve`, I get a 404 error for the developer notes.

We need to build the site first: run `uv run mkdocs build`
Sorry I should have included a check list/instruction for reviewers.

.pre-commit-config.yaml
```diff
@@ -21,6 +21,7 @@ repos:
     rev: 0.7.22
     hooks:
       - id: mdformat
+        exclude: ^docs/(index|devnotes|api/index)\.md$
```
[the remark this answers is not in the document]
tanhaow: These files need to be excluded because When mdformat runs on these files, it auto-corrects the snippet scissors they contain from "8<" to "8/<", which breaks the MkDocs snippet inclusion feature.

.pre-commit-config.yaml
```diff
@@ -31,6 +32,7 @@ repos:
     rev: v4.5.0
     hooks:
       - id: check-yaml
+        exclude: ^mkdocs\.yml$
```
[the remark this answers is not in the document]
tanhaow: This exclusion is also for skipping auto-correction for MkDocs's Snippets extension that we are using. Snippets needs to set a relative path with [specific YAML tags](https://github.com/Princeton-CDH/remarx/blob/enhance-mkdocs/mkdocs.yml#L59C7-L59C41) that standard YAML formatter don't recognize. I'll add a comment explaining it.

src/remarx/__init__.py
```diff
@@ -3,4 +3,8 @@
 # Copyright (c) 2025, Center for Digital Humanities, Princeton University
 # SPDX-License-Identifier: Apache-2.0
 
+from remarx import app, sentence
+
 __version__ = "0.1.0.dev0"
+
+__all__ = ["__version__", "app", "sentence"]
```
laurejt, line 10: We don't include `app_utils` here, but for some reason it is being displayed... Is this the behavior we expect?

.pre-commit-config.yaml
```diff
@@ -21,6 +21,9 @@ repos:
     rev: 0.7.22
     hooks:
       - id: mdformat
+        # Exclude docs that use MkDocs snippet scissors (--8<--) since mdformat
+        # converts them to --8/<-- which breaks the snippet inclusion feature
+        exclude: ^docs/(index|devnotes|api/index)\.md$
```
laurejt, line 26: Why is `api/index.md` being excluded? It doesn't contain any snippets...

laurejt (changes requested):
Overall, this looks pretty good. I have a few requested changes / outstanding questions:

- Update the mkdocs options with the suggestions I've made to improve the look of the code documentation page.
- How can we control what modules are being displayed? Despite not being added to the top-level `__init__.py`, `app_utils` is still being displayed in the code documentation page.
- As possible, remove the exclusions from the pre-commit hooks. Instead of removing the files, see if there's a way to configure the tools to ignore the particularly issue.

On a related note, it looks like we should be able to add a top-level docstring to `app.py` without marimo removing it. I'll go ahead and add one in #176

rlskoeser (commented):
Revisions look good, and I'm glad that this approach allows us to simplify the docs. 

My concerns are aligned with @laurejt's - if possible, let's configure our formatters to work with our content rather than ignoring. If it's too much effort, then documenting clearly as you have done will make it possible to revisit in future.

tanhaow:
I’ve made changes based on the review. Thank you @laurejt for suggesting better configurations for mkdocs!

The markdown and yaml formatters now have configured to work with our content rather than ignore it:
1. **check-yaml** now allows unknown custom tags with `args: [--unsafe]` (so we can use `!relative` tag in `mkdocs.yml`)
2. **mdformat** now tolerate our snippet scissors with additional plugins. (thank you @rlskoeser !)

tanhaow:
@laurejt, thank you for bringing this to my attention. I spent some time trying to fix it. It’s actually quite a tricky problem.

Apart from removing the imports logic from `corpus/__init__.py` (which I suppose we don’t want to do), to resolve the duplicated entries issue, we’ll have to manually create the [API index page](https://github.com/Princeton-CDH/remarx/blob/11a79f96b992eb2cda0703190a9c02efac7bd3cd/docs/api/index.md) to have more granular control. Our `mdformat` hook would try to formatting the `options` section so for now I have to temporally exclude the index page from `mdformat` check.

image
Currently the TOC looks good. There's no more duplicates. It's just the code gets less elegant.

tanhaow:
> Almost there, but please fix the cross-reference error for `TEIinput`. It causes the following generation error.
> 
> image

It seems because cross-reference is a paid feature in mkdocstrings... I removed cross reference for now.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -145,9 +145,8 @@ class TEIinput(FileInput):
 
     def __post_init__(self) -> None:
         """
-        After default initialization, parse [input_file][remarx.sentence.corpus.text_input.TextInput.input_file]
-        as a [TEIDocument][remarx.sentence.corpus.tei_input.TEIDocument] and store as
-        [xml_doc][remarx.sentence.corpus.tei_input.TEIinput.xml_doc].
+        After default initialization, parse the input file as a TEIDocument
+        and store it as xml_doc.
```
laurejt, line 149: I guess we'll just need to look into why the crossref is failing later

laurejt (approved):
Seems good for now. Hopefully we can circle back to the code documentation indexing and crossref issues in the near future.

rlskoeser:
> It seems because [cross reference](https://mkdocstrings.github.io/python/usage/configuration/docstrings/?h=cross+re#relative_crossrefs) is a paid feature in mkdocstrings... I removed cross reference for now.

Regular cross-references work, it's relative and scoped cross-references that are paid features (the relative links would be a lot nicer and shorter).  

The reason this stopped working is that the classes I was referencing were not included in the imports and thus no longer included anywhere in the documentation. I've restored the cross-references and fixed the inputs so the classes are present to be linked to.

Issue #161, opened 2025-08-08
Add GitHub Action to check changelog


laurejt:
See: https://github.com/marketplace/actions/check-changelog-for-changes

PR #173, opened 2025-08-19
Create Check Changelog Github Action

tanhaow:
**Associated Issue(s):** #161 

### Changes in this PR
This pull request introduces a new GitHub Actions workflow (`.github/workflows/changelog.yml`) to automatically [check for changelog updates](https://github.com/marketplace/actions/check-changelog-for-changes) on pull requests targeting the `develop` and `main` branches. The workflow helps ensure that changes are properly documented in `CHANGELOG.md` before merging.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,8 @@
 name: Changelog Updated
 on:
   pull_request:
+    # opened/synchronize/reopened: Check changelog when PR is created or updated
+    # labeled/unlabeled: Re-run check when "no changelog" label is added/removed
     types: [opened, synchronize, reopened, labeled, unlabeled]
     branches: [develop, main]
 jobs:
--- a/DEVELOPERNOTES.md
+++ b/DEVELOPERNOTES.md
@@ -46,6 +46,21 @@ The `CHANGELOG.md` is meant for end users and should document user-facing change
 Internal changes like CI/CD updates, build system modifications, or development tooling
 changes should not be included in the changelog unless it is substantial enough to potentially impact functionality, such as a major refactor.
 
+### To Skipping Changelog Check for Specific PRs
+
+Not all changes require a changelog entry. For PRs that contain non-user-visible changes (such as CI/CD updates, internal refactoring, documentation fixes, or development tooling changes), you can skip the changelog requirement:
+
+1. **Add the `no changelog` label** to your PR:
+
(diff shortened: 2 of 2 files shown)


.github/workflows/changelog.yml
```diff
@@ -0,0 +1,14 @@
+name: Check Changelog
```
rlskoeser, line 1: Let's rename the workflow so that it is more meaningful when we see it on the list of checks for a pull request.

```suggestion
name: Changelog Updated
```

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,14 @@
...
+on:
+  pull_request:
+    types: [assigned, opened, synchronize, reopened, labeled, unlabeled]
```
rlskoeser, line 4: Do we need this? Are there any excluded or non-default types here?

pyproject.toml
```diff
@@ -61,6 +61,7 @@ testpaths = ["tests"]
 
 [tool.coverage.run]
 branch = true
+omit = ["src/remarx/app.py", "docs/hooks.py"]
```
rlskoeser, line 64: What is the `docs/hooks.py` you're excluding here? I don't see it in the file list, is this from another branch?

.github/workflows/changelog.yml
```diff
@@ -0,0 +1,14 @@
...
+    branches: [develop, main]
+jobs:
+  check-changelog:
+    name: Check Changelog Action
```
rlskoeser, line 8: This name doesn't seem very meaningful and does display on the check badge on the PR, but I don't have any suggestions to rename...

rlskoeser (changes requested):
Thanks for setting this up. Looks reasonable. 

A few minor things:
- I'm suggesting a workflow name change
- Please omit PR types unless needed; if needed, please add a comment briefly noting why
- Remove the coverage omit path 
- Document somewhere (developer notes?) how to exempt a PR from this check, since there will occasionally be PRs that don't require updates to the change log.

pyproject.toml
laurejt, line 1: This file should not need to be modified. This PR is not connected to the mkdocs effort.

laurejt (changes requested):
Looks pretty good. In addition to @rlskoeser 's comments, we should revisit as a team what (if any) development setup should be included within the change log.

rlskoeser (commented):
Thanks for revising. This can be merged once the documentation and comment changes are made, I don't need to review again.

PR #181, opened 2025-08-22
Feature/tei seperate footnotes

tanhaow:
**Associated Issue(s):** #

### Changes in this PR
_Include all key changes in this pull request_

- Change 1
- Change 2

### Notes
_Include any additional notes that will help in the reviewing of this pull request_

- Note 1
- Note 2

### Reviewer Checklist
_Include **discrete** checks that should be done by the reviewer beyond looking through
code and/or file changes. Note that this check list will correspond to tasks within
the PR overview page._

- [ ] Check 1
- [ ] Check 2

### DIFF
--- a/src/remarx/sentence/corpus/tei_input.py
+++ b/src/remarx/sentence/corpus/tei_input.py
@@ -15,7 +15,7 @@ from typing import ClassVar, NamedTuple, Self
 from lxml.etree import XMLSyntaxError
 from neuxml import xmlmap
 
-from remarx.sentence.corpus.text_input import FileInput
+from remarx.sentence.corpus.base_input import FileInput
 
 TEI_NAMESPACE = "http://www.tei-c.org/ns/1.0"
 
@@ -53,14 +53,17 @@ class TEIPage(BaseTEIXmlObject):
     text_nodes = xmlmap.StringListField("following::text()")
     "list of all text nodes following this tag"
 
-    def text_contents(self) -> Generator[str]:
+    def text_contents(self) -> Generator[tuple[str, str]]:
         """
         Generator of text content on this page, between the current
-        and following page begin tags.  MEGA specific logic:
-        ignores page indicators for the manuscript edition
+        and following page begin tags. Enhanced to return 2 chunks per page:
+        body text separate from footnotes, with section type information.
+        MEGA specific logic: ignores page indicators for the manuscript edition
         (<pb> tags with ed="manuscript"); assumes standard pb tags have no edition.
+        Note: Footnotes that span multiple pages are excluded for now.
         """
-        # for now, ignore partial content, footnotes, hyphenation, etc
+        # Extract body text (ignore partial content, hyphenation, etc, and skip footnotes)
+        body_text_parts = []
         for text in self.text_nodes:
             # text here is an lxml smart string, which preserves context
             # in the xml tree and is associated with a parent tag.
@@ -74,24 +77,55 @@ class TEIPage(BaseTEIXmlObject):
             ):
                 break
 
-            # omit editorial content (e.g. original page numbers)
-            if (
-                parent.tag == TEI_TAG.add
-                or (parent.tag == TEI_TAG.label and parent.get("type") == "mpb")
-            ) and (text.is_text or (text.is_tail and text.strip() == "")):
-                # omit if text is inside an editorial tag (is_text)
-                # OR if text comes immediately after (is_tail) and is whitespace only
-                continue
+            # Skip this text node if it's inside a footnote
+            ancestor = parent
+            while ancestor is not None:
+                if ancestor.tag == TEI_TAG.note and ancestor.get("type") == "footnote":
+                    break
+                ancestor = ancestor.getparent()
+            else:
+                # omit editorial content (e.g. original page numbers)
+                if (
+                    parent.tag == TEI_TAG.add
+                    or (parent.tag == TEI_TAG.label and parent.get("type") == "mpb")
+                ) and (text.is_text or (text.is_tail and text.strip() == "")):
+                    # omit if text is inside an editorial tag (is_text)
+                    # OR if text comes immediately after (is_tail) and is whitespace only
+                    continue
 
-            # consolidate whitespace if it includes a newline
-            # (i.e., space between indented tags in the XML)
-            yield re.sub(r"\s*\n\s*", "\n", text)
+                # consolidate whitespace if it includes a newline
+                # (i.e., space between indented tags in the XML)
+                body_text_parts.append(re.sub(r"\s*\n\s*", "\n", text))
+
+        # Yield body text as first chunk
+        body_text = "".join(body_text_parts).strip()
+        if body_text:
+            yield (body_text, "text")
+
+        # Extract and yield combined footnotes as second chunk
+        notes = self.node.xpath(
+            "following::t:note[@type='footnote'][not(.//t:pb[not(@ed)])]",
+            namespaces=self.ROOT_NAMESPACES,
+        )
+
+        if notes:
+            footnote_texts = []
+            for note in notes:
+                footnote_text = "".join(
+                    note.xpath(".//text()", namespaces=self.ROOT_NAMESPACES)
+                ).strip()
+                if footnote_text:
+                    footnote_texts.append(footnote_text)
+
+            if footnote_texts:
+                yield ("\n\n".join(footnote_texts), "footnote")
 
     def __str__(self) -> str:
         """
         Page text contents as a string
         """
-        return "".join(self.text_contents())
+        # Extract just the text content from the tuples
+        return "".join(text for text, section_type in self.text_contents())
 
 
 class TEIDocument(BaseTEIXmlObject):
@@ -137,7 +171,11 @@ class TEIinput(FileInput):
     xml_doc: TEIDocument = field(init=False)
     "Parsed XML document; initialized from inherited input_file"
 
-    field_names: ClassVar[tuple[str, ...]] = (*FileInput.field_names, "page_number")
+    field_names: ClassVar[tuple[str, ...]] = (
+        *FileInput.field_names,
+        "page_number",
+        "section_type",
+    )
     "List of field names for sentences from TEI XML input files"
 
     file_type = ".xml"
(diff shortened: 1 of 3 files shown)


laurejt:
Shouldn't the change log be modified, so we can document how TEI processing differs from text processing?

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -74,24 +78,57 @@ def text_contents(self) -> Generator[str]:
...
+                    footnote_text = re.sub(r"\s*\n\s*", "\n", footnote_text)
+                    footnote_text_parts.append(footnote_text)
+
+            if footnote_text_parts:
+                yield ("\n\n".join(footnote_text_parts), "footnote")
 
     def __str__(self) -> str:
         """
-        Page text contents as a string
+        Page body text contents as a string
```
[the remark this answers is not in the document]
tanhaow: Thanks for the pointer! Do we want the __str__ function to return both body text and footnotes, or just return the body text?

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -53,14 +53,18 @@ class TEIPage(BaseTEIXmlObject):
...
+    def text_contents(self) -> Generator[tuple[str, str]]:
         """
         Generator of text content on this page, between the current
-        and following page begin tags.  MEGA specific logic:
-        ignores page indicators for the manuscript edition
+        and following page begin tags. Enhanced to return 2 chunks per page:
+        body text separate from footnotes, with section type information.
+        MEGA specific logic: ignores page indicators for the manuscript edition
         (<pb> tags with ed="manuscript"); assumes standard pb tags have no edition.
+        Note: Footnotes that span multiple pages are excluded for now.
```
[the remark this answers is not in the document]
laurejt: This is fair, I was just puzzled by their being a unit test for this case.

laurejt:
> > Shouldn't the change log be modified, so we can document how TEI processing differs from text processing?
> 
> @laurejt I didn't update changelog because TEI processing has been mentioned in @rlskoeser's [earlier PR](https://github.com/Princeton-CDH/remarx/pull/157/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed:~:text=Add%20%60corpus%60%20submodule%20with%20input%20classes%20for%20text%20and%20TEI/XML), and I felt this PR is an update facing developers, not directly to users.

I guess that's fair. Perhaps what I'm asking for makes less sense. I think it depends on how much we want to flag that this is a preliminary version in the change log, which very well could be overkill.

tanhaow:
> Thanks for digging into this and making progress.
> 
> Requested changes:
> 
> * Split out logic for body and footnote content into separate methods OR separate modes of one method (I'm now thinking this might be simpler; then the string method can call it to get all text content)
> * Move logic for checking if content is inside a footnote into a method and write unit tests for it
> * Correct the footnote logic so that we only return footnotes on the current page. Update your test fixture so you include multiple pages that have footnotes and confirm only footnotes for the current page are included (and that all of them are included for their respective page).
> * Rather than using strings for section type everywhere, let's declare them as a string enum (`StrEnum`) in the base input class.

Thanks for your good suggestions. I've made the following changes:
1. Split out logic for body and footnote content into separate methods; having the str dunder method to get all text content.
2. Correct the footnote logic to only return footnotes on the current page.
3. Declare section type using `StrEnum` <-- great advice! Thank you!

This PR is ready for review again. @rlskoeser @laurejt

laurejt (changes requested):
Thanks for making the additional changes.

Requested changes:
- Fix `TEIPage`'s `__str__` method so sections are not combined with whitespace separation
- Update documentation to explicitly state that each footnote is yielded individually
- Update unit tests so that they are properly scoped to the method being tested and do note test the behavior of the methods used by the tested method.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -53,22 +54,71 @@ class TEIPage(BaseTEIXmlObject):
...
-    def text_contents(self) -> Generator[tuple[str, str]]:
+    # fetch all footnotes after the current page break; will filter them in Python later
+    # pb is a delimiter (not a container), so "following::note" returns all later footnotes
+    all_following_footnotes = xmlmap.NodeListField(
+        "following::t:note[@type='footnote']", xmlmap.XmlObject
+    )
+    "list of all footnote elements following this page"
+
+    @staticmethod
+    def _is_footnote_content(el: etree._Element) -> bool:
```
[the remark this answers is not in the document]
tanhaow: Originally it was just because it's an internal method that's only used within the TEIPage class itself.
But I agree that this helper method could potentially be useful for other classes that needs to determine if an element is footnote content. I can make it public

rlskoeser (changes requested):
Great progress.  

I've made several comments in line with suggestions and requests for improving and simplifying the code, and to make better use of neuxml functionality.

In addition to the changes I've requested in my comments, please also:
- Update from develop so this code can be tested with a working create script (How did you test it when working on it?)
- Rename and update `TEIPage` text and footnote methods and properties so they are consistent. I think properties should be both named with `following_` (but `all` seems redundant); methods should be named similarly, and should return the same type (both string rather than one string and one generator)

tanhaow:
The implementation now addresses the review comments received and is ready for review again.
The only remaining item is a low-priority performance optimization (to bail out early at some level) in `is_footnote_content()` that we can address in a future enhancement if needed.

laurejt (changes requested):
Thanks for the updates.

Request Changes:
- There is redundant (perhaps residual) logic for generating footnote text. I believe both `get_footnote_contents` and `_get_footnote_contents_generator` should be removed (and the `TEIinput` class subsequently updated).
- Reorganize unit tests so that `get_body_text` and `get_footnote_text` are tested separately. Both scenarios can be tested within the same unit test.
- Further refine unit tests to mock functions fully.
  - When mock methods are used, the mocks should be tested for being called the correct number of times and what inputs they were called with (as appropriate)
  - If a method relies on other methods within the codebase, make sure to mock them as possible to avoid having a failure cascade across several unit tests. For example, `test_str` should have patched/mocked methods. This unit test shouldn't fail if `get_body_text` or `get_footnote_text` is broken.

tanhaow:
Hi @laurejt, thanks for the feedback! I've revised some parts of the code based on your review.

Regarding your question about why `get_body_text` and `get_footnote_contents` weren't patched/mocked in `test_get_text()`, and why I didn't patch `get_text` in `test_get_sentences()`: I was following @rlskoeser 's original testing style before my added footnote feature ([see here](https://github.com/Princeton-CDH/remarx/blob/release/0.1/tests/test_sentence/test_corpus/test_tei_input.py#L106C4-L123C13)), which uses sample TEI files and only mocks external dependencies (like segment_text). I tried to maintain this consistency when adding the footnote functionality.

But I also totally understand your point about better isolation and prevent cascading failures. Question: Which testing approach should I adopt going forward? Continue to use sample TEI files or refactor to mocking for isolation?
Happy to refactor based on your preference!

tests/test_sentence/test_corpus/test_tei_input.py
```diff
@@ -52,22 +54,69 @@ def test_attributes(self):
...
         # first text content after the pb tag
         assert text.startswith("als in der ersten Darstellung.")  # codespell:ignore
         # last text content after the next standard pb tag
         assert text.endswith("entwickelten nur das Bild der eignen Zukunft!")
         # should not include editorial content
         assert "|" not in text
         assert "IX" not in text
-        # TODO: eventually should not include footnote content
+
+    def test_get_body_text_with_footnotes(self):
```
rlskoeser, line 93: I was confused by the method name. Suggest adding a comment to clarify:
```suggestion
    def test_get_body_text_with_footnotes(self):
         # test a sample page with footnotes to confirm footnote contents are excluded
```

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -83,15 +128,43 @@ def text_contents(self) -> Generator[str]:
...
+        """
+        footnotes = self.get_page_footnotes()
+        if not footnotes:
+            return ""
+        # consolidate whitespace once after joining all footnotes
+        return re.sub(
+            r"\s*\n\s*",
+            "\n",
+            "\n\n".join(str(footnote).strip() for footnote in footnotes),
+        )
```
rlskoeser, line 161: I expect this regex to remove your double newline; I don't think you need the regex substitution here, just do the join.  Your unit test for this method should check that the expected newline is present between footnotes.

Is a double newline needed or would one be sufficient?

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -154,12 +231,26 @@ def __post_init__(self) -> None:
...
         for page in self.xml_doc.pages:
-            yield {"text": str(page), "page_number": page.number}
+            body_text = page.get_body_text()
+            if body_text:
+                yield {
+                    "text": body_text,
+                    "page_number": page.number,
+                    "section_type": SectionType.TEXT.value,
+                }
+
```
rlskoeser, line 250: Please add a comment here with a note explaining why we're yielding individual footnotes rather than one chunk of footnote content.  (I missed this convo & decision, but also seems important to remember for future revisions.)

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -83,15 +128,43 @@ def text_contents(self) -> Generator[str]:
...
+        Get all footnote content as a single string, with footnotes separated by double newlines.
+        """
+        footnotes = self.get_page_footnotes()
+        if not footnotes:
+            return ""
+        # consolidate whitespace once after joining all footnotes
+        return re.sub(
+            r"\s*\n\s*",
+            "\n",
+            "\n\n".join(str(footnote).strip() for footnote in footnotes),
```
rlskoeser, line 160: Please update to use `get_individual_footnotes` method. No need to bail out early, if the list is empty join will return an empty string.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -83,15 +128,43 @@ def text_contents(self) -> Generator[str]:
...
+        )
 
     def __str__(self) -> str:
         """
-        Page text contents as a string
+        Page text contents as a string, with body text and footnotes.
         """
-        return "".join(self.text_contents())
+        return f"{self.get_body_text()}\n\n{self.get_footnote_text()}".strip()
```
rlskoeser, line 167: Do you really need to call `strip()` here?

rlskoeser (changes requested):
Thanks for your work on this, it's much closer and I think I should be able to review the remaining revisions fairly quickly. 

Here are the changes I'd like made before this is ready to merge:
- Simplify the test for end of page in `get_body_text`
- Use the same method to generate footnote content in `__str__` and the text yielded to generate the sentence corpus; I think this means using `get_individual_footnotes` both places; if `get_footnote_text`  is only used by the string method maybe it should be dropped (although the parallel with body text is nice!); if you keep it, it should be simplified and the delimiter and whitespace logic must be tested
- Add a comment documenting why we're yielding individual footnotes 
- Improve unit tests so that the core logic of each method is directly tested

PR #183, opened 2025-08-22
Specify download reuse for stanza resources for sentence segmentation

rlskoeser:
**Associated Issue:**  ~ 

### Changes in this PR

- Configure sentence segment method to reuse stanza resources instead of checking every time on pipeline init

### Notes

- This is to partially address the slowness we're seeing when segmenting the full MEGA TEI
- I could have sworn we had this in the first version of the segmentation implementation! Must have gotten lost when we simplified on the refactor.

### DIFF
--- a/src/remarx/sentence/segment.py
+++ b/src/remarx/sentence/segment.py
@@ -5,6 +5,7 @@ sentence begins and the sentence text itself.
 """
 
 import stanza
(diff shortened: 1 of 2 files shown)


tanhaow (approved):
Look good to me 🚀

Issue #163, opened 2025-08-08
Create separate ruff linting/formatting job in GitHub Actions workflow


tanhaow:
Currently, our GitHub Actions workflow runs a unit test and code coverage job. 
We also want to add a separate job that runs **ruff linting and formatting**.
Since linting doesn’t depend on test results, and vice versa, the jobs should run in parallel.

PR #187, opened 2025-08-26
Add GitHub actions to check mkdocs build and ruff style

rlskoeser:
**Associated Issue(s):** #163 

### Changes in this PR

- Set minimum uv version in pyproject.toml and remove from unit test workflow
- Simplified unit test workflow (python setup is included in uv setup)
- Add new workflow to check docs and ruff formatting
- Shorten unit test workflow name and add status badge to readme

### Notes

- After Laure caught a cross-reference that was causing a problem, I thought it would be important to have an automated check to flag these for us. This is adapted in part from a similar check on my undate project.
- I've configured the check workflow to run on PRs and manually only, thought that would be enough; let me know if you think we should add/modify triggers.
- No review checklist items other than reviewing the code changes

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [develop, main]
   # Allow running manually
-  on: workflow_dispatch
+  workflow_dispatch:
 jobs:
   check_docs:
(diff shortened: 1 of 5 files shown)


tanhaow (approved):
Good to merge.

Issue #174, opened 2025-08-20
make a 0.1 release branch


mnaydan:
- [x] document how you prepped the release branch
- [x] review changelog and see if anything needs to be consolidated/organized
- [x] revise the software release checklist for this project (app specific things in developer notes)

_Reminder: the release branch should include the issues delivered for acceptance testing and any in progress work merges to develop._

PR #188, opened 2025-08-27
Release/0.1

tanhaow:
**Associated Issue(s):** #174


rlskoeser:
@tanhaow @laurejt Something about the change log was bothering me so I went looking for guidelines and found this https://common-changelog.org/

I've made some revisions to the change log, please review when you get the chance.

Issue #122, opened 2025-08-04
As a user, I want to get a CSV of sentences in my text documents, created by selecting one file of the supported types so that I can create a sentence corpus that can feed into the core pipeline.


mnaydan:
## Testing Notes (QA):

Follow the instructions in the `remarx` app.

- [x] Confirm you can upload a plaintext file (`.txt`) and generate a sentence corpus (`.csv`).
- [x] Confirm you cannot upload an unsupported file type (e.g a PDF file) through drag and drop or selection. 
- [x] Confirm you can open the generated sentence corpus file in a spreadsheet application.
- [x] Confirm the generated sentence corpus file location matches what is reported by the app.
- [x] Check that the file field within the generated CSV matches the input text's filename.
- [x] Check that the sentence index field (`sent_index`) is present and increases sequentially throughout the whole CSV file.

--- 
- first version plaintext only
- Should warn / error if you choose an unsupported file type
- When running locally: download not needed, but report where the file(s) were created
- Choose a simple default for output folder for now so we can put off configurability

-----
- [x] App with custom logic for building sentence corpora
  -  Calls `segment_text` directly
  - Only accepts a single plaintext files as input
- [x] Modify app to use `TextInput` instead (note this was never officially reviewed / merged to develop)
- [x] Modify app to rely on some kind of "corpus.create" method that relies on the input factory

laurejt:
Will now include integrating corpus building classes

bn2409-svg:
Works great!

PR #189, opened 2025-08-27
Update base output fields so sentence index is included in sentence csv

rlskoeser:
**Associated Issue:** #122

### Changes in this PR

- Update base input class with the correct field names (replace `offset` with `sent_index`)
- Update unit tests to match the change


laurejt (approved):
🚀

Issue #215, opened 2025-09-16
Add basic sentence ID logic to corpus construction


laurejt:
## Testing notes

Open the sentence corpus file (`.CSV`) you just created in #132 using a TEL-XML file.
- [x] Confirm that the output CSV file contains the `sent_id` column.
- [x] Confirm that the values in the `sent_id` column are sequential and follow the `filename:number` format, such as `MEGA_A2_B005-00_ETX.xml:10`.
- [x] Confirm that each row in the output CSV contains a unique sentence id in the `sent_id` column.

Now, go to the app and do the same corpus building routine, but this time using a `TXT` file, such as [1895-96a.txt](https://drive.google.com/file/d/1MPeUiAxAtBsxm5KkONF__aixITy9MGv0/view?usp=drive_link).

- [x] Confirm that the output CSV file contains the `sent_id` column.
- [x] Confirm that the values in the `sent_id` column are sequential and follow the `filename:number` format, such as `1895-96a.txt:10`.
- [x] Confirm that each row in the output CSV contains a unique sentence id in the `sent_id` column.

---

Update `base_input.get_sentences` so that it also includes a sentence id.

PR #217, opened 2025-09-16
Feature/add sentence ids

laurejt:
**Associated Issue(s):** #215

### Changes in this PR

- Adds sentence id field to a `FileInput` generated sentence objects
- Added new unit test for `FileInput.get_sentences`
- Updated `field_names` unit tests for all input classes. Updated the children classes so that they are only testing the added fields (and therefore won't fail if we add another field to the base class)

### Notes
_Include any additional notes that will help in the reviewing of this pull request_

- Right now I've defined the sentence ids to be a combination of the file name and sentence index: file:sent_index. I'm open to other ideas, but this seemed like it'd work for now (there's some potential for trouble if there end up being multiple input files with the same name)

### Reviewer Checklist

@rlskoeser 
- [x] Running the app locally, verify that created sentence corpora include sentence ids for both file types
- [x] Verify that the resulting sentence ids are correct

@tanhaow 
- [x] Running the app locally, verify that created sentence corpora include sentence ids for both file types
- [x] Verify that the resulting sentence ids are correct

### DIFF
--- a/src/remarx/sentence/corpus/base_input.py
+++ b/src/remarx/sentence/corpus/base_input.py
@@ -66,8 +66,9 @@ class FileInput:
 
         :returns: Generator of one dictionary per sentence; dictionary
         always includes: `text` (text content), `file` (filename),
-        `sent_index` (sentence index within the document). It may include
-        other metadata, depending on the input file type.
+        `sent_index` (sentence index within the document), and `sent_id`
+        (sentence id). It may include other metadata, depending
+        on the input file type.
         """
         # zero-based sentence index for this file, across all chunks
         sentence_index = 0


tests/test_sentence/test_corpus/test_text_input.py
```diff
@@ -19,7 +20,7 @@ def test_file_name(tmp_path: pathlib.Path):
 
 
 def test_field_names(tmp_path: pathlib.Path):
-    assert TextInput.field_names == ("file", "sent_index", "text")
+    assert TextInput.field_names == FileInput.field_names
```
tanhaow, line 23: Great!

Issue #209, opened 2025-09-15
Generate sentence embeddings from an sbert-accessible german language model


mnaydan:
Create a method for generating sentence embeddings. Perhaps called `get_sentence_embeddings`.

**Input.**`list[str]`: list of sentences

**Returns.** `npt.NDArray` (numpy n-dimensional array) : The extracted *normalized* embeddings for the sentences in the input corpus. This is a 2-dimensional array of size n_sentences x n_dimensions.

**Reference.** `extract_sentence_embeddings` in [get_sentence_embeddings.py](https://github.com/Princeton-CDH/remarx/blob/experiments/embeddings/src/remarx/get_sentence_embeddings.py)

**Pretrained model.** `paraphrase-multilingual-mpnet-base-v2`

PR #218, opened 2025-09-17
Generate sentence embeddings from an sbert-accessible german language model

tanhaow:
**Associated Issue(s):** #209

### Changes in this PR

- Adds `get_sentence_embeddings` method for generating sentence embeddings from CSV sentence corpora
- Adds `validate_sentence_corpus` helper function to validate CSV file structure and required columns
- Added unit tests covering validation logic for both
- Added dependency group `embeddings` to manage embedding-related dependencies
- Updated `sentence/__init__.py` to include the new embeddings module

### Notes
- I created a helper function to validate if the input CSV file adheres to the structure of our corpus building process. This function might also be useful for other features in the future. But I’m unsure of the appropriate location to place it since we don’t have a `util` file. 
- The model’s progress bar is always enabled to log in the console. But I couldn't found a way to make it possible to be integrated it with Marimo.

### Reviewer Checklist @laurejt 

- [ ] Generate a sentence corpus CSV file with the latest corpus building process (to make sure it has all the required columns)
- [ ] Run `uv sync` and then `remarx-generate-embeddings your_corpus.csv(input) embeddings.npy(output)`
- [ ] Check if the code is good enough and passes all unit tests

### Reviewer Checklist @rlskoeser 
- [ ] Generate a sentence corpus CSV file with the latest corpus building process (to make sure it has all the required columns)
- [ ] Run `uv sync` and then `remarx-generate-embeddings your_corpus.csv(input) embeddings.npy(output)`
- [ ] Check if the code is good enough and passes all unit tests

### DIFF
--- a/tests/test_quotation/test_embeddings.py
+++ b/tests/test_quotation/test_embeddings.py
@@ -4,31 +4,12 @@ Tests for sentence embedding functionality.
 
 from unittest.mock import Mock, patch
 
-import pytest
-
 from remarx.quotation.embeddings import get_sentence_embeddings
 
 
-@pytest.mark.parametrize(
-    "sentences,model_name,expected_model",
-    [
-        (
-            ["Test sentence 1", "Test sentence 2"],
-            None,
-            "paraphrase-multilingual-mpnet-base-v2",
-        ),
-        (
-            ["Test sentence 1", "Test sentence 2"],
-            "paraphrase-multilingual-mpnet-base-v3",
-            "paraphrase-multilingual-mpnet-base-v3",
-        ),
-    ],
-)
 @patch("remarx.quotation.embeddings.SentenceTransformer")
-def test_get_sentence_embeddings(
-    mock_transformer_class, sentences, model_name, expected_model
-):
-    """Test sentence embedding generation with different models and inputs."""
+def test_get_sentence_embeddings(mock_transformer_class):
+    """Test sentence embedding generation from list of sentences."""
 
     # Mock the sentence transformer
     mock_model = Mock()
@@ -36,14 +17,14 @@ def test_get_sentence_embeddings(
     mock_model.encode.return_value = mock_embeddings
     mock_transformer_class.return_value = mock_model
 
-    # Call function with or without model_name
-    if model_name is None:
-        result = get_sentence_embeddings(sentences)
-    else:
-        result = get_sentence_embeddings(sentences, model_name=model_name)
+    sentences = ["Test sentence 1", "Test sentence 2"]
 
-    # Verify the model was initialized with expected model name
-    mock_transformer_class.assert_called_once_with(expected_model)
+    result = get_sentence_embeddings(sentences)
+
+    # Verify the model was initialized with default model name
+    mock_transformer_class.assert_called_once_with(
+        "paraphrase-multilingual-mpnet-base-v2"
+    )
 
     # Verify encode was called with correct parameters
     mock_model.encode.assert_called_once_with(
@@ -52,4 +33,14 @@ def test_get_sentence_embeddings(
         show_progress_bar=False,
     )
 
+    assert result == mock_embeddings
+
+    # Test with custom model
+    mock_transformer_class.reset_mock()
+    custom_model = "paraphrase-multilingual-mpnet-base-v3"
+
+    result = get_sentence_embeddings(sentences, model_name=custom_model)
+
+    # Verify custom model was used
+    mock_transformer_class.assert_called_once_with(custom_model)
     assert result == mock_embeddings


pyproject.toml
```diff
@@ -49,13 +51,20 @@ docs = [
   "mkdocstrings[python]",
 ]
 html = [ "bs4" ]
+embeddings = [
+  "sentence-transformers",
+  "torch",
+  "transformers",
+  "ipython",  # Required by transformers for trainer functionality
```
laurejt, line 58:
```suggestion
  "transformers[torch]",
```

pyproject.toml
```diff
@@ -49,13 +51,20 @@ docs = [
   "mkdocstrings[python]",
 ]
 html = [ "bs4" ]
+embeddings = [
```
laurejt, line 54: Seems to me like this should also be included in `test`. More generally, it's not clear to me that should be separated from the dependencies (but this is more subjective).

laurejt (changes requested):
This is a minimal review due to the design and scope changes for this issue. Below are some of the high-level changes, **but these are non-exhaustive since I have not fully reviewed this PR.**

- Remove the project command and CLI interface from the embedding generation
- Update the generate sentence embeddings method so it takes a list of sentences (`list[str]`). As a result, remove all file-related code including the file validation.
- Unit tests should not mock at the package / module level.

tanhaow:
@laurejt Thanks for the feedback! `get_sentence_embeddings()` now takes `list[str]` instead of file paths. Ready for another look when you have time!

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,37 @@
+"""
+Provides functionality to generate sentence embeddings from sentence corpora
+using pretrained models from the sentence-transformers library.
+"""
```
laurejt, line 4: Seems like we could probably reduce this to something like this
```suggestion
Library for generating sentence embeddings from pretrained Sentence Transformer models.
```

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,37 @@
...
+
+import numpy.typing as npt
+
+try:
+    from sentence_transformers import SentenceTransformer
+except ImportError as e:
+    raise ImportError(
+        "The sentence-transformers library is required for embedding functionality. "
+        "Install it by running: uv sync"
+    ) from e
```
laurejt, line 14: There should be no situation where the package does not include the sentence-transformers package. Please remove.

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def get_sentence_embeddings(
+    sentences: list[str],
+    model_name: str = "paraphrase-multilingual-mpnet-base-v2",
+) -> npt.NDArray:
+    """
+    Extract sentence embeddings for each sentence in the input list using the specified pretrained Sentence
+    Transformers model and return them as a 2-dimensional numpy array.
```
laurejt, line 23: Current description is insufficient. It should include:
1. a description of the optional parameter and its default
2. describe the shape of the array

For more detail, see the docstring for the  [method](https://github.com/Princeton-CDH/remarx/blob/experiments/embeddings/src/remarx/get_sentence_embeddings.py#L27) from the embedding experiments.

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,37 @@
...
+    :param sentences: List of sentences to generate embeddings for
+    :param model_name: Name of the pretrained sentence transformer model to use (leave as default for German)
+    :return: 2-dimensional numpy array of normalized sentence embeddings
+    """
+    # Generate embeddings using the specified model
+    model = SentenceTransformer(model_name)
+    embeddings = model.encode(
+        sentences,
+        normalize_embeddings=True,
+        show_progress_bar=True,  # Output progress bar in console
```
laurejt, line 34: This should also be an optional parameter.

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+from unittest.mock import Mock, patch
+
+import numpy as np
+import pytest
+
+import remarx.quotation.embeddings
+from remarx.quotation.embeddings import get_sentence_embeddings
+
+
+class TestGetSentenceEmbeddings:
```
laurejt, line 15: This should not be a class, because the thing being tested is not a class

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+Tests for sentence embedding functionality.
+"""
+
+import importlib
+from unittest.mock import Mock, patch
+
+import numpy as np
+import pytest
+
+import remarx.quotation.embeddings
```
laurejt, line 11: Remove, this is unnecessary

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+
+def test_sentence_transformers_import_error():
+    """Test that ImportError is raised when sentence-transformers is not available."""
+
+    with (
+        patch.dict("sys.modules", {"sentence_transformers": None}),
+        pytest.raises(ImportError, match="sentence-transformers library is required"),
+    ):
+        # Force reimport to trigger the ImportError
+        importlib.reload(remarx.quotation.embeddings)
```
laurejt, line 83: Delete

pyproject.toml
```diff
@@ -37,18 +38,25 @@ dev = [
...
   "coverage[toml]", # Enables coverage to read config from pyproject.toml
+  "remarx[embeddings]",
 ]
 docs = [
   "mkdocs",
   "mkdocs-material",
   "mkdocstrings[python]",
 ]
 html = [ "bs4" ]
+embeddings = [
```
laurejt, line 55: Correction: This is not optional. These are dependencies of the core package.

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+        np.testing.assert_array_equal(result, mock_embeddings)
+
+    @patch("remarx.quotation.embeddings.SentenceTransformer")
+    def test_get_sentence_embeddings_custom_model(self, mock_transformer_class):
+        """Test sentence embedding generation with custom model."""
+
+        # Mock the sentence transformer
+        mock_model = Mock()
+        mock_embeddings = np.array([[0.1, 0.2]])
+        mock_model.encode.return_value = mock_embeddings
```
laurejt, line 54: Given that there's no validation of this output, just make this a string so that the numpy dependency can be removed from this unit test.

E.g., 
```suggestion
        mock_model.encode.return_value = "embeddings"
```

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,83 @@
...
+        # Mock the sentence transformer
+        mock_model = Mock()
+        mock_embeddings = np.array([[0.1, 0.2, 0.3], [0.4, 0.5, 0.6]])
+        mock_model.encode.return_value = mock_embeddings
+        mock_transformer_class.return_value = mock_model
+
+        sentences = [
+            "Das ist ein Test.",  # codespell:ignore
+            "Dies ist ein weiterer Test.",  # codespell:ignore
+        ]
```
laurejt, line 29: No reason to include german sentences here given that they are not being processed.

tests/test_quotation/test_embeddings.py
laurejt, line 1: If the value and type of a mock class do not matter for the purposes of testing, keep them simple. For example, the numpy dependency is not needed for these unit tests.

src/remarx/quotation/embeddings.py
tanhaow: Yeah I saw that in experiment you made `show_progress_bar` optional, but I was thinking we may always want to see the progress in console, so I simplified this option. But happy to add it back.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 - Add sentence id field (`sent_id`) to generated sentence corpora
 - Processes TEI/XML documents to yield separate chunks for body text and footnotes, with each footnote yielded individually as a separate element
+- Add a method for generating sentence embeddings from sentence corpus
```
rlskoeser, line 17: Suggesting an edit since the method now doesn't take a corpus file directly
```suggestion
- Add a method for generating sentence embeddings from a list of sentences
```

src/remarx/quotation/embeddings.py
laurejt: @rlskoeser please correct me if I'm wrong, but generally we want to have a way of disabling this for testing.

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,34 @@
...
```
rlskoeser, line 31: I'm curious, how does progress work? Any chance we'll be able to integrate a marimo progressbar in future? (out of scope for now, though!)

src/remarx/quotation/embeddings.py
rlskoeser: agreed, would be useful to control progressbar behavior when we call the method for running in marimo

rlskoeser (commented):
I've taken a quick look but haven't pulled to test anything locally.

I agree with Laure - any dependencies that are required for the core functionality of the application should be included in the main dependency list and you can assume they are installed in the code.

src/remarx/quotation/embeddings.py
tanhaow: The progress comes from tqdm deep inside sentence-transformers' batch processing, so it's not easily accessible. We don’t get a clean callback/handle :(

pyproject.toml
```diff
@@ -22,12 +22,16 @@ classifiers = [
   "Topic :: Text Processing",
 ]
 dependencies = [
+  "ipython",         # Required by transformers for trainer functionality
```
laurejt, line 25: Remove this is not needed, the package manager knows to install the transformers dependency.

src/remarx/quotation/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+This module contains functionality for embedding generation and quotation detection.
```
laurejt, line 2: Might as well use similar language as the`sentence` module
```suggestion
This module contains libraries for embedding generation and quotation detection.
```

src/remarx/quotation/embeddings.py
```diff
@@ -0,0 +1,29 @@
...
+    :param sentences: List of sentences to generate embeddings for
+    :param model_name: Name of the pretrained sentence transformer model to use (default: paraphrase-multilingual-mpnet-base-v2)
+    :return: 2-dimensional numpy array of normalized sentence embeddings with shape [# sents, # dims]
+    """
+    # Generate embeddings using the specified model
+    model = SentenceTransformer(model_name)
+    embeddings = model.encode(
+        sentences,
+        normalize_embeddings=True,
+        show_progress_bar=False,  # default to False to disable progress bar output in unit tests
```
laurejt, line 27: This isn't an optional argument anymore, so it's not default so much as always off (I think this is the default behavior of `encode`).

If the progress bar option is being removed entirely, don't specify this optional parameter.

tests/test_quotation/test_embeddings.py
```diff
@@ -0,0 +1,64 @@
...
+        sentences,
+        normalize_embeddings=True,
+        show_progress_bar=False,
+    )
+
+    assert result == mock_embeddings
+
+
+@patch("remarx.quotation.embeddings.SentenceTransformer")
+def test_get_sentence_embeddings_custom_model(mock_transformer_class):
```
laurejt, line 40: There's no reason to have this as a separate test, which will let you remove all the duplicate code.

laurejt (changes requested):
This is looking pretty good, I've got a few small requested changes; it should be good to go after this.

- Remove the ipython dependency (we don't need to explicitly mention it)
- Refactor the unit tests into a single test to reduce code duplication. Mocks can be reused, just reset them to get the correct call logic (it won't impact return values).

pyproject.toml
tanhaow: after removing this it won't pass the unit tests in Github CI

src/remarx/quotation/embeddings.py
tanhaow: I meant to change it in input parameter

pyproject.toml
tanhaow: I just removed it and tested again. Here's the error: `ModuleNotFoundError: No module named 'IPython'`

Issue #211, opened 2025-09-15
As a user, I want a way to access log files so that I can troubleshoot application behavior.


jerielizabeth:
## Testing notes

Run `uv run remarx-app` to launch the app (if it is not open).

- [x] Confirm there's a path for log files displayed at the bottom of the `Sentence Corpus Builder` page.
- [x] Confirm there's a path for log files displayed at the bottom of the `Quote Finder` page.
- [x] Find the log file on your local computer, and check that it includes something like the following:
```
[2025-10-01 14:48:21] INFO:remarx-app::Remarx application starting
[2025-10-01 14:48:21] INFO:remarx-app::Logs are being written to: [your local path]
[2025-10-01 14:48:22] INFO:remarx-app::Remarx Corpus Builder notebook started
[2025-10-01 14:48:23] INFO:remarx-app::Remarx Quote Finder notebook started
```
- [x] Upload your log file in the comment section at the bottom of this issue.

---

## Dev notes
Use python logging to configure a basic default; listen to remarx at some log level (warn or info) and output logging to a file that can be found by users to share if they have trouble - maybe a logs directory relative to where they are running it?

Add a notice somewhere in the app notebook to report where the log file is.

As possible, incorporate stanza logging. If this isn't possible, add an option for silencing stanza.

bn2409-svg:
Looks great! Here's the log: 

[2025-10-04 11:04:14] INFO:remarx-app::Remarx application starting
[2025-10-04 11:04:14] INFO:remarx-app::Logs are being written to: /Users/bennettnightingale/logs/remarx_20251004_110414.log
[2025-10-04 11:04:16] INFO:remarx-app::Remarx Corpus Builder notebook started
[2025-10-04 11:20:28] INFO:remarx-app::Remarx Corpus Builder notebook started
[2025-10-04 11:20:34] INFO:remarx-app::Remarx Quote Finder notebook started
[2025-10-04 11:20:34] INFO:remarx-app::Remarx Quote Finder notebook started
[2025-10-04 11:26:45] INFO:remarx-app::Remarx Corpus Builder notebook started
[2025-10-04 11:46:10] INFO:remarx-app::Remarx Quote Finder notebook started
[2025-10-04 11:48:31] INFO:remarx.quotation.pairs::Now generating sentence embeddings
[2025-10-04 11:48:31] INFO:sentence_transformers.SentenceTransformer::Use pytorch device_name: mps
[...]

Issue #166, opened 2025-08-08
Set up logging


laurejt:
- Set up logging for the package
- Potentially allow option to suppress stanza's logging. See: [](https://stanfordnlp.github.io/stanza/getting_started.html#controlling-logging-from-the-pipeline)

laurejt:
Forgotten predecessor of #211

PR #220, opened 2025-09-18
Configure logging and suppress stanza

tanhaow:
**Associated Issue(s):** #211 #166 

### Changes in this PR
- Added logging configuration in `utils.py` with `setup_logging()` function (set to INFO level)
- Integrated logging setup into application startup flow via `launch_app()` function
- Display log file location Marimo notebook UI
- Configured automatic creation of timestamped log files in `./logs/` directory
- Suppressed Stanza's logging to ERROR level

### Reviewer Checklist
- [x] Run `remarx-app` and verify log file is created in `./logs/` directory with correct timestamp format
- [x] Check that log file location is displayed in the Marimo notebook UI
- [x] Confirm that Stanza loggings are suppressed from log files and console

### DIFF
--- a/src/remarx/utils.py
+++ b/src/remarx/utils.py
@@ -2,6 +2,7 @@
 Utility functions for the remarx package
 """
 
+import io
 import logging
 import pathlib
 from datetime import datetime
@@ -20,12 +21,12 @@ def configure_logging(
     :param log_destination: Where to write logs. Can be:
         - None (default): Creates a timestamped log file in ./logs/ directory
         - pathlib.Path: Write to the specified file path
-        - Any text stream (e.g., sys.stdout, sys.stderr, or any io.TextIOBase): Write to the given stream
+        - Any io.TextIOBase (e.g., sys.stdout, sys.stderr, or any io.TextIOBase): Write to the given stream
     :param log_level: Logging level for remarx logger (default to logging.INFO)
     :param stanza_log_level: Logging level for stanza logger (default to logging.ERROR)
     :return: Path to the created log file if file logging is used, None if stream logging
     """
-    # Determine logging configuration based on log_destination parameter
+
     log_file_path: pathlib.Path | None = None
     config_output_opts: dict
     if log_destination is None:
@@ -35,8 +36,8 @@ def configure_logging(
         timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
         log_file_path = log_dir / f"remarx_{timestamp}.log"
         config_output_opts = {"filename": log_file_path, "encoding": "utf-8"}
-    elif hasattr(log_destination, "write"):
-        # Stream logging to any text stream (e.g., sys.stdout, sys.stderr, or any TextIO)
+    elif isinstance(log_destination, io.TextIOBase):
+        # Only allow io.TextIOBase instances as streams (includes sys.stdout, sys.stderr)
         config_output_opts = {"stream": log_destination}
     else:
         # File logging to specified path
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -55,7 +55,7 @@ def test_configure_logging_default_creates_timestamped_filename(tmp_path, monkey
 def test_configure_logging_stream(tmp_path, monkeypatch, stream):
     # Run in a temporary CWD and ensure no logs/ directory is created when streaming to a text stream
     monkeypatch.chdir(tmp_path)
-    created_path = configure_logging(stream, log_level=logging.INFO)
+    created_path = configure_logging(log_destination=stream, log_level=logging.INFO)
 
     assert created_path is None
 


rlskoeser (changes requested):
This looks great, and does pretty nearly meet the needs of the story as written, which was specific to the marimo app. I'd like to expand the scope slightly so we can use the same logging for development and command line scripts in future.  (We can increase estimate if appropriate.)

Here's a summary of the changes I'm requesting:
- move the method to a top-level remarx.utils and rename (setup -> configure)
- take optional parameters for filename / stream and log level; defaults are what you've implemented already
- import and call the method in the notebook rather than passing filename by environment variable
- move log file info text to the bottom of the notebook
- write a unit test for the logging method to confirm default behavior and parameter handling

src/remarx/app/ui.py
```diff
@@ -45,7 +45,25 @@ def _(mo):
...
-    mo.md(rf"""Running `remarx` version: {remarx.__version__}""")
+    import os
+    import logging
+
+    # Get log file path
+    log_file = os.environ.get('REMARX_LOG_FILE')
+
+    if log_file:
+        # Log that UI started
+        logger = logging.getLogger(__name__)
```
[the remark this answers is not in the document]
laurejt: Found the relevant logging code within marimo: https://github.com/marimo-team/marimo/blob/4c17a0edf644f963ff7e4931e3dad7402efc1b8b/marimo/_loggers.py#L92

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    logging.getLogger().info("Configuring logging with default file works")
+
+    # Flush the log file (required by ruff check)
+    for handler in logging.getLogger().handlers:
+        with contextlib.suppress(Exception):
+            handler.flush()
+
+    assert "Configuring logging with default file works" in created_path.read_text(
+        encoding="utf-8"
+    )
```
rlskoeser, line 48: I don't think this part of the test is necessary.

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    assert isinstance(created_path, Path)
+    assert created_path.parent.name == "logs"
+
+    # Check that the log file name starts with "remarx_" and ends with ".log"
+    assert created_path.name.startswith("remarx_")
+    assert created_path.suffix == ".log"
+    assert created_path.exists()
+
+    # Stanza logger level default
+    assert logging.getLogger("stanza").getEffectiveLevel() == logging.ERROR
```
rlskoeser, line 36: Can you do the same thing for a different context to check that it is the expected level? 

Or can you use logic similar to what you're doing in the `reset_logging` fixture to inspect the configured logging handler?

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_configure_logging_default_creates_timestamped_file(tmp_path, monkeypatch):
+    """Test that the default configuration creates a timestamped log file."""
+    # Run in a temporary CWD so logs land under tmp_path/logs/
+    monkeypatch.chdir(tmp_path)
+    created_path = configure_logging()
+
+    assert isinstance(created_path, Path)
+    assert created_path.parent.name == "logs"
```
rlskoeser, line 28: Could be helpful for readability to explicitly  check that `tmp_path / "logs"` is a directory.

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@pytest.fixture(autouse=True)
+def reset_logging():
+    """Ensure a clean logging configuration for each test."""
+    root_logger = logging.getLogger()
+    for handler in root_logger.handlers[:]:
+        root_logger.removeHandler(handler)
+        with contextlib.suppress(Exception):
+            handler.close()
```
rlskoeser, line 18: Interesting, would not have thought about this! Without this were you getting an accumulation of logging handlers?

I wondered if any of the built-in pytest fixtures would be helpful... it seems that `caplog` is really only for inspecting logging.

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+    created_path = configure_logging(sys.stdout, log_level=logging.INFO)
+
+    assert created_path is None
+
+    logging.getLogger().info("Configuring logging with stdout works")
+
+    # Capture the output and check that it contains the log message
+    captured = capsys.readouterr()
+
+    assert "Configuring logging with stdout works" in captured.out
```
rlskoeser, line 61: If there's a way to inspect the logging handler and check that it's configured as expected, that would be better than generating output and confirming it's present. Alternately, you could use a mock and inspect the call args.

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def test_configure_logging_specific_file(tmp_path):
+    target = tmp_path / "nested" / "custom.log"
+    created_path = configure_logging(target, log_level=logging.DEBUG)
+
+    assert created_path == target
+    assert target.exists()
```
rlskoeser, line 69: Should the created file exist? Or just the parent directory?

tests/test_utils.py
```diff
@@ -0,0 +1,94 @@
...
+
+    assert "Configuring logging with default file works" in created_path.read_text(
+        encoding="utf-8"
+    )
+
+
+def test_configure_logging_stdout_stream(capsys):
+    created_path = configure_logging(sys.stdout, log_level=logging.INFO)
+
+    assert created_path is None
```
rlskoeser, line 54: Maybe worth using temp path and confirming log dir is not created?

src/remarx/utils.py
```diff
@@ -0,0 +1,73 @@
...
+        log_file_path = log_dir / f"remarx_{timestamp}.log"
+
+        logging.basicConfig(
+            filename=log_file_path,
+            level=log_level,
+            format="[%(asctime)s] %(levelname)s:%(name)s::%(message)s",
+            datefmt="%Y-%m-%d %H:%M:%S",
+            encoding="utf-8",
+        )
+        created_log_file = log_file_path
```
rlskoeser, line 44: The `created_log_file` variable is redundant and also potentially confusingly named if a path is passed in. I suggest setting `log_file_path = None` before the if statements and using that for both logging configuration and return value.

src/remarx/utils.py
```diff
@@ -0,0 +1,73 @@
...
+        log_file_path = log_destination
+        log_file_path.parent.mkdir(parents=True, exist_ok=True)
+
+        logging.basicConfig(
+            filename=log_file_path,
+            level=log_level,
+            format="[%(asctime)s] %(levelname)s:%(name)s::%(message)s",
+            datefmt="%Y-%m-%d %H:%M:%S",
+            encoding="utf-8",
+        )
```
rlskoeser, line 67: Let's consolidate the three calls to `logging.basicConfig` - only one set of parameters is changing, so you can create a dict to set the filename or stream option (maybe name something like `config_output_opts`) and then you pass it into the `basicConfig` method call as keyword args with `**` notation.

src/remarx/utils.py
```diff
@@ -0,0 +1,73 @@
...
+    Configure logging for the remarx application.
+    Supports logging to stdout, a specified file, or auto-generated timestamped file.
+
+    :param log_destination: Where to write logs. Can be:
+        - None (default): Creates a timestamped log file in ./logs/ directory
+        - pathlib.Path: Write to the specified file path
+        - sys.stdout: Write to console output
+    :param log_level: Logging level for remarx logger (default to logging.INFO)
+    :param stanza_log_level: Logging level for stanza logger (default to logging.ERROR)
+    :return: Path to the created log file if file logging is used, None if stream logging
```
rlskoeser, line 27: This all looks great.

rlskoeser (changes requested):
Revisions look great! This will be very helpful to get merged.

Your unit tests look more like integration tests, since you're writing log messages and then checking for them. We can trust that python logging will work properly. We only need to test that we have configured the log handler the way we intend.

I'm requesting some small changes; hopefully not too difficult:
- Refactor `configure_logging` to remove the extra `created_log_file` variable and only call `basicConfig()` once
- Revise unit tests to remove the logging calls and reading log files; either inspect the configured log handler and log levels to confirm logging is configured as expected, or if that is difficult, use a mock to confirm the logging config is called as expected.

tests/test_utils.py
tanhaow: Yes — without this reset fixture, handlers were accumulating across tests (because pytest runs them in one process). `caplog` is good for asserting on logs, but it also doesn’t isolate logging configuration. One way to get rid of this fixture is to call `logging.basicConfig(..., force=True)` to replace existing handlers instead of stacking them. Do you want me to switch to that?

Issue #225, opened 2025-09-22
Refactor the app into multiple notebooks


tanhaow:
## Testing notes

Run `uv run remarx-app` to launch the app.

- [ ] Confirm that you are able to navigate to the `Quote Finder` page via the navigation bar at the top of the page.

---

- [x] Undo the accordion layout and put the content back 
- [x] split into two notebooks (build corpus & detect quotation)
- [x] a script to combine and launch two notebooks 
- [x] minimal navigation to switch between the two notebooks.

Useful link:
https://docs.marimo.io/guides/deploying/programmatically/#fastapi-example

PR #228, opened 2025-09-29
Feature/refactor app

tanhaow:
**Associated Issue(s):** #225

### Changes in this PR
- [x]  Undo the accordion layout and put the content back
- [x]  split into two notebooks (build corpus & detect quotation)
- [x]  a script to combine and launch two notebooks
- [x]  minimal navigation to switch between the two notebooks.


src/remarx/app/utils.py
```diff
@@ -7,19 +7,55 @@
...
+    """Launch the remarx app in the default web browser"""
+    # Create marimo asgi app
+    server = mo.create_asgi_app()
+
+    # Add notebooks
+    ## For now set corpus builder as landing page
+    server = server.with_app(path="", root=remarx.app.corpus_builder.__file__)
+    server = server.with_app(
+        path="/corpus-builder", root=remarx.app.corpus_builder.__file__
+    )
```
rlskoeser, line 30: Rather than bind the notebook twice, let's implement the root url as a redirect to the corpus-builder url. We can always add a top level landing page later if we decide to add one.

src/remarx/app/utils.py
```diff
@@ -7,19 +7,55 @@
...
+    return mo.vstack(
+        [
+            mo.md("# `remarx`").center(),
+            mo.md(f"Running version: {remarx.__version__}").center(),
+            mo.md("---"),
+            mo.nav_menu(
+                {
+                    "/corpus-builder": "## Sentence Corpus Builder",
+                    "/quote-finder": "## Quote Finder",
+                }
```
rlskoeser, line 54: It would be nice if this reflected state - i.e., currently open app should be highlighted but not clickable (clicking results in a reload, which could lose state).

It doesn't look like nav menu supports this, so we probably should call this good enough for the initial refactor.

src/remarx/app/utils.py
```diff
@@ -7,19 +7,55 @@
...
+    server = server.with_app(
+        path="/quote-finder", root=remarx.app.quote_finder.__file__
+    )
+
+    # Create a FastAPI app
+    app = FastAPI()
+    app.mount("/", server.build())
+
+    # Run server
+    uvicorn.run(app, host="localhost", port=8000)
```
rlskoeser, line 40: The script still needs a bit of code to automatically opening the server url in the browser.

This can be done with the python `webbrowser.open`, but it can't be called after `uvicorn.run` because that doesn't return until the app is shut down.

We can implement that behavior with a lifespan context manager https://fastapi.tiangolo.com/advanced/events/#lifespan 

... but I can't find any way that the context manager can determine what the host and url are, which probably means we have to hard-code them somewhere in the code where both context manager and uvicorn.run can use them.  I guess they're already hard-coded so it maybe isn't such a big deal, but ideally in future they might be configurable.

rlskoeser: Sorry I didn't share example code! Here was my experimental version of the above when I was testing if it would work:
```python

    # Run server
    host = "localhost"
    port = 8000

    @asynccontextmanager
    async def lifespan(app: FastAPI) -> None:
        webbrowser.open(f"http://{host}:{port}/")
        yield

    # Create a FastAPI app
    app = FastAPI(lifespan=lifespan)
    app.mount("/", server.build())

    uvicorn.run(app, host=host, port=port)
   
```
This requires two additional imports:
```python
import webbrowser
from contextlib import asynccontextmanager
```


I tested putting the context manager inside the `launch_app` method and outside it. It would feel better and be easier to test it if it is a separate method, so let's go with that and just set the host and port at the file level so both methods can use the same variables.

tanhaow: It works perfectly! Thank you for discovering this approach (I didn't know this `webbrowser.open` built-in method before!).
I'll fix the unit tests, and this branch should be able to merge soon.

tests/test_app/test_ui.py
```diff
@@ -1,6 +1,13 @@
...
-    app.run()
+
+def test_corpus_builder_app():
+    assert hasattr(corpus_builder, "app")
+    assert isinstance(corpus_builder.app, marimo.App)
+
+
+def test_quote_finder_app():
+    assert hasattr(quote_finder, "app")
+    assert isinstance(quote_finder.app, marimo.App)
```
rlskoeser, line 13: What is this testing?

tests/test_app/test_app_utils.py
```diff
@@ -28,6 +50,43 @@ def test_create_temp_input(mock_upload):
...
+async def test_lifespan(mock_webbrowser):
+    """Test lifespan context manager"""
+    from fastapi import FastAPI
+
+    # Create a mock app
+    app = FastAPI()
+
+    # Test the lifespan context manager
+    async with lifespan(app):
+        mock_webbrowser.open.assert_called_once_with("http://localhost:8000/")
```
rlskoeser, line 83: interesting syntax! I haven't used python async stuff yet.

Briefly made me wonder if we could call it like this in the app and pass in host and port, but I see now this is for testing.

src/remarx/app/utils.py
```diff
@@ -5,21 +5,78 @@
...
+@asynccontextmanager
+async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
+    """Lifespan context manager to open browser when server starts"""
+    webbrowser.open(f"http://{HOST}:{PORT}/")
+    yield
+
+
+async def redirect_root() -> RedirectResponse:
+    """Redirect root path to corpus-builder"""
+    return RedirectResponse(url="/corpus-builder", status_code=302)
```
rlskoeser, line 36: I thought there'd be a more efficient way to declare this in the app setup, but seems fine. We might want to tweak the comment for future us
```suggestion
async def redirect_root() -> RedirectResponse:
    """Redirect root path to corpus-builder, since app currently has no home page"""
    return RedirectResponse(url="/corpus-builder", status_code=302)
```

rlskoeser (commented):
This runs locally and code looks fine.  My only question is what the unit tests for the notebook apps/pages are accomplishing and whether that's the best we can do.

tests/test_app/test_ui.py
tanhaow: This simply test if they are `marimo.App` instances - I know they are quite useless... but without them the codecov report won't pass. Do you have suggestions on how we can make this test better?

laurejt (changes requested):
The app is looking good, it ran fine for me locally. I wonder if there's a way for us to disable the `nav_menu` when the corpus builder (and in the future, quote detection), but that's a future problem.

Requested changes:
- The notebook unit tests should run the notebook programmatically (see the original test `test_app` )
- The logging for both notebooks should be tested. See the [unit test](https://github.com/Princeton-CDH/remarx/blob/feature/detect-quotes/tests/test_app/test_ui.py#L9) I wrote in a separate branch.

laurejt:
Another thing we need to add: logging for the second notebook.

tests/test_app/test_ui.py
tanhaow: I see. I was wondering what kinds of assertion test we can use here. But seems like simply using `run` should be enough (if it crushes, the test fails.)

laurejt:
Another larger issue with logging: multiple log files are created if there's any navigation; this is not the behavior we want.

tanhaow:
> Another larger issue with logging: multiple log files are created if there's any navigation; this is not the behavior we want.

Thank you for catching this. I'll look into how to fix this.

tanhaow:
> > > Another larger issue with logging: multiple log files are created if there's any navigation; this is not the behavior we want.
> > 
> > 
> > Thank you for catching this. I'll look into how to fix this.
> 
> I wonder if we could create some kind of singleton pattern, but I'm not sure how we'd save the state since the notebooks are run from scratch each time they are launched.

The multiple log files issue can be fixed by let the notebooks always write into the same log file instead of creating a new log file every time it launches. Do we want this implementation? It actually might be easier for users to report the errors since there will only be one log file.

rlskoeser:
> We can create a global variable to store the current log file path to prevent the creation of multiple log files. Alternatively, we can reduce the granularity of the timestamped log files. Currently, they are created based on `datetime`, but we can change it to `date` instead. This way, the app can reuse existing log files on the same day if available.
> 
> @rlskoeser, what are your thoughts on this?

Would you try moving the configure the logging call to the launch script instead of the notebooks? My expectation / hope is that this would create one log shared by both notebooks for the duration of the session.

If that doesn't work, reducing the granularity of the timestamp and ensuring logging is appended seems like a good solution. Anther option could be a logfile with a standard name and no timestamp, and configuring a rotating log file handler to rename on some period / size, but that seems more complex than the first two options.

tanhaow:
> > We can create a global variable to store the current log file path to prevent the creation of multiple log files. Alternatively, we can reduce the granularity of the timestamped log files. Currently, they are created based on `datetime`, but we can change it to `date` instead. This way, the app can reuse existing log files on the same day if available.
> > @rlskoeser, what are your thoughts on this?
> 
> Would you try moving the configure the logging call to the launch script instead of the notebooks? My expectation / hope is that this would create one log shared by both notebooks for the duration of the session.
> 
> If that doesn't work, reducing the granularity of the timestamp and ensuring logging is appended seems like a good solution. Anther option could be a logfile with a standard name and no timestamp, and configuring a rotating log file handler to rename on some period / size, but that seems more complex than the first two options.

The first one seems like a great approach. I'll implement it and see if it solves the problem. Thanks for your suggestions!

tanhaow:
@laurejt thank you for helping make the final changes. Let's merge and close this PR.

Issue #210, opened 2025-09-15
Use voyager to find sentence pairs


mnaydan:
Method that identifies likely quote original-reuse sentence pairs for a given original and reuse sentence corpus and their associated sentence embeddings.

**Inputs.**
- `pathlib.Path`: reuse sentence corpus
- `pathlib.Path`: original sentence corpus
- `pathlib.Path`: file path for output CSV

NOTE: This method will call the sentence embedding creation function.

**Side Effects.** Saves identified quote sentence-pair information as CSV.

**Implementation Details.**
- Use approximate nearest neighbors to identify likely quote sentence-pairs.
- **For now, only fetch the nearest neighbor.** (This perhaps make too strong of assumptions)
- Q: Use some low threshold to filter out bad pairs?
- Use dot product (i.e., cosine similarity) for "distance", use this value as the "match score"

**Output File Format.** See the [technical design document](https://github.com/Princeton-CDH/remarx/blob/develop/docs/technical-design-document.md#quote-sentence-pairs) for details.

**Reference.** [annoy notebook](https://github.com/Princeton-CDH/remarx/blob/experiments/embeddings/notebooks/try-annoy.py)

PR #231, opened 2025-09-30
Revise detect quotes code to use voyager in place of annoy

rlskoeser:
**Associated Issue(s):** #210

### Changes in this PR

- Library for detecting sentence-level quote pairs pairs.py using voyager
- Updates change log, adds new section quote detection
- Updated pyproject.toml to include restrict polars versions (due to read_csv)
- Added logging to pairs.py and embeddings.py and added unit testing for logging

---

### Changes in this PR

- This is a revision to feature/detect-quotes branch that replaces annoy with voyager
   - switched cutoff to  `> 0.2 ` instead of `< 0.8` since we get a distance instead of a similarity score
- I started working on revising the unit tests but they are not fully working with these changes; there are some open questions it would be good to resolve first

### Notes

- I've tried both inner product and cosine for the space configuration but I get the same distance scores in the resulting output. Not sure if it's something I'm doing or something in the library.
- Wondered about switching from cosine similarity to cosine distance since that's what we get out of the system.


rlskoeser:
@laurejt this PR is currently pointing at your annoy branch as the base so you can see the differences, but I don't think we should worry about reconciling them (as they've started to diverge). Once you've reviewed the changes against your branch let's switch the PR to develop.

src/remarx/quotation/pairs.py
```diff
@@ -169,40 +154,41 @@ def find_quote_pairs(
     original_corpus: pathlib.Path,
     reuse_corpus: pathlib.Path,
     out_csv: pathlib.Path,
-    score_cutoff: float = 0.8,
+    score_cutoff: float = 0.2,
```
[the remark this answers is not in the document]
laurejt: That seems reasonable to me.

src/remarx/quotation/pairs.py
```diff
@@ -77,43 +74,31 @@ def get_sentence_pairs(
...
-    matches = []
-    progress_bar = tqdm(
-        enumerate(reuse_vecs),
-        desc="Finding sentence pairs",
-        total=reuse_vecs.shape[0],
-        disable=not show_progress_bar,
-    )
+    # Get sentence matches; query all vectors at once
+    # returns a list of lists with results for each reuse vector
+    all_neighbor_ids, all_distances = index.query(reuse_vecs, k=1)
```
laurejt, line 88: it seems like we should log how long this takes.

src/remarx/quotation/pairs.py
```diff
@@ -77,43 +74,31 @@ def get_sentence_pairs(
...
-                }
-            )
-    return pl.DataFrame(matches)
+        # add row index
+        .with_row_index(name="reuse_index")
+        # since we requested k=1, explode the lists to get single value result
+        .explode("original_index", "match_score")
+        # then filter by specified match score cutoff
+        .filter(pl.col("match_score").lt(score_cutoff))
+    )
```
laurejt, line 100: Perhaps we should have the logging here for how many matches are found

src/remarx/quotation/pairs.py
```diff
@@ -77,43 +74,31 @@ def get_sentence_pairs(
...
+    # Build search index
     # NOTE: An index only needs to be generated once for a set of embeddings.
     #       Perhaps there's some potential reuse between runs?
     if logger.isEnabledFor(logging.INFO):
         start = time()
-    index = build_annoy_index(original_vecs, n_trees)
+    index = build_vector_index(original_vecs, n_trees)
     if logger.isEnabledFor(logging.INFO):
         elapsed_time = time() - start
         logger.info(f"Built Annoy index in {elapsed_time:.1f} seconds")
```
laurejt: Perhaps we should move this logging to `build_vector_index`?

src/remarx/quotation/pairs.py
laurejt, line 88: We should consider giving access to the following voyager parameters:
- `query_ef`: this will impact the quality of the results
- `num_threads`: if we want to allow multi-threading

src/remarx/quotation/pairs.py
```diff
@@ -169,40 +154,41 @@ def find_quote_pairs(
...
     )
-    if logger.isEnabledFor(logging.INFO):
-        elapsed_time = time() - start
-        logger.info(
-            f"Identified {len(sent_pairs)} sentence pairs in {elapsed_time:.1f} seconds"
-        )
+    elapsed_time = time() - start
+    logger.info(
+        f"Identified {len(sent_pairs)} sentence pairs in {elapsed_time:.1f} seconds"
+    )
```
laurejt, line 182: The more I look at this, the more it seems like this logging should be within `get_sentence_pairs`

src/remarx/quotation/pairs.py
```diff
@@ -169,40 +154,41 @@ def find_quote_pairs(
...
-    if logger.isEnabledFor(logging.INFO):
+    # Build and save quote pairs if any are found
+    if len(sent_pairs):
+        quote_pairs = compile_quote_pairs(original_df, reuse_df, sent_pairs)
+        # NOTE: Perhaps this should return a DataFrame rather than creating a CSV?
+        quote_pairs.write_csv(out_csv)
         logger.info(f"Saved {len(quote_pairs)} quote pairs to {out_csv}")
+    else:
+        logger.info(
+            "No sentence pairs for specified score cutoff; output file not created."
```
laurejt, line 192: It seems worth including the cutoff value within the log.

rlskoeser (commented):
@laurejt I think this looks good and is ready to merge!

@tanhaow do you want to test running this version locally?

tanhaow:
> @laurejt I think this looks good and is ready to merge!
> 
> @tanhaow do you want to test running this version locally?

I'm running it right now. Will report back in a sec.

tanhaow:
@laurejt Thank you for finding out the solution to this tricky problem. The result looks great. I think it's good to merge!

Issue #212, opened 2025-09-15
Generate output with sentence pairs (text and any initial sentence corpus metadata)


mnaydan:
## Testing notes

Open the generated sentence pairs file you just created in the last issue (#213) in a spreadsheet application:

- [ ] Confirm the generate file contains the following columns: `reuse_id`, `reuse_text`, `original_id`, `original_text`, `match_score`

Note for @rlskoeser : We should upload a sample output file here.

---

**NOTE:** This issue is blocked by #210

Revise the quote detection logic so that:
- All original and reuse corpus fields are included within the output CSV
~- Add quote ids to the pairs~

(This issue description will be revised as #210's development progresses)

PR #233, opened 2025-10-01
Add additional corpus metadata to quote pairs CSV 

laurejt:
**Associated Issue(s):** #212 

### Changes in this PR

- Updates `pairs.py` logic so it now carries any additional corpus metadata to the output CSV
- Some small updates to the docstrings in `pairs.py` so they render properly for mkdocs
- Updated technical design doc to reflect current development

### Notes

- The ordering of the fields for the output field have changed. Now, the ordering resembles the ordering of the final output within the technical design doc.

### Reviewer Checklist

@rlskoeser 
- [x] Check that the script runs locally.
- [x] Verify that the resulting output CSV has the additional fields and that the values of these fields have been preserved.

@tanhaow 
- [ ] Check that the script runs locally.
- [ ] Verify that the resulting output CSV has the additional fields and that the values of these fields have been preserved.

### DIFF
--- a/src/remarx/quotation/pairs.py
+++ b/src/remarx/quotation/pairs.py
@@ -49,9 +49,9 @@ def get_sentence_pairs(
     sentence pairs where quotation is likely. Returns these sentence pairs as
     a polars DataFrame including for each pair:
 
-        - original_index : the index of the original sentence
-        - reuse_index : the index of the reuse sentence
-        - match_score: the quality of the match
+    - `original_index`: the index of the original sentence
+    - `reuse_index`: the index of the reuse sentence
+    - `match_score`: the quality of the match
 
     Likely quote pairs are identified through the sentences' embeddings. The Annoy
     library is used to find the nearest original sentence for each reuse sentence.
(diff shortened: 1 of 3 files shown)


src/remarx/quotation/pairs.py
```diff
@@ -106,52 +106,53 @@ def get_sentence_pairs(
...
+    [load_sent_df][remarx.quotation.pairs.load_sent_df].
 
     Returns a dataframe with the following fields:
-        - reuse_id: ID of the reuse sentence
-        - original_id: ID of the original sentence
-        - match_score: Estimated quality of the match
+
+    - `match_score`: Estimated quality of the match
+    - All non-index fields from the reuse corpus (`reuse_id`, `reuse_text`, ...)
+    - All non-index fields from the original corpus (`original_id`, `original_text`, ...)
```
rlskoeser, line 149: I'm not sure what you mean by "non-index" here, could you clarify / simplify? Do you just mean everything but the row index?  (we also have sentence indexes)

src/remarx/quotation/pairs.py
```diff
@@ -106,52 +106,53 @@ def get_sentence_pairs(
...
     """
     # Build and return quote pairs
     return (
         detected_pairs.join(reuse_corpus, on="reuse_index")
         .join(original_corpus, on="original_index")
-        .select(
-            ["reuse_id", "reuse_text", "original_id", "original_text", "match_score"]
-        )
+        .drop(["reuse_index", "original_index"])
     )
```
rlskoeser, line 156: I checked against the technical design doc in develop and it doesn't look to me that this is following the proposed order there. I don't think it's a blocker but wanted to check if there's anything you're intentionally wanting to modify about field order at this point, and if so would you update the technical design doc as part of this PR.

rlskoeser (commented):
Looks good, works well locally. I just have a question about field order and a suggestion for a slightly more streamlined way to re-order the polars columns.

tanhaow (approved):
Sorry for getting this back to you late. I was working on the release this afternoon.
The code looks good to me. I tested it locally and it works well.

PR #234, opened 2025-10-01
connect find_quote_pairs to marimo notebook

tanhaow:
**Associated Issue(s):** #

- Connected `find_quote_pairs` to `Quote Finder` notebook (currently using the default threshold 0.225)
- Add necessary widgets and instructions to let user can actually run the method (referenced Corpus Builder)

### Reviewer Checklist
- [ ] Test the method using the interface.
- [ ] Check the code.


rlskoeser (commented):
Works well locally! I think we may want to revise some of the language in the notebook at some point, but hopefully we can get project team input on that.

Issue #221, opened 2025-09-19
Run passim on the MEGA texts in comparison with the DNZ texts 


mnaydan:
- [x] Look at corppa code and Laure's notes
- [x] Look at passim documentation and examples
- [x] Output as json file
- [ ] Document how you did it in case we need to rerun

References:
- [Passim notes](https://docs.google.com/document/d/1Vh8CeABjp5FvTEZo1SEeqcJXwIPfbSzAfK_1XY7XJgQ/edit?tab=t.0) with relevant notes on the citing marx project
- [Notes on passim parameters](https://docs.google.com/document/d/1x5ke_IoW_p5YAaHGlhvYasPOa07o4llVoTLcwGiONBg/edit?tab=t.0#heading=h.ydbwsd2yay57) catering to the PPA-NLP project
- [passim-related corppa code](https://github.com/Princeton-CDH/corppa/tree/develop/src/corppa/poetry_detection/passim)

PR #245, opened 2025-10-14
Script/passim results

tanhaow:
**Associated Issue(s):** #221 

### Changes in this PR

Scripts:
- `convert_xml.py`: Script to convert MEGA XML files into plain text files, organized by page.
- `convert_texts_to_passim_corpus.py`: Script to convert text files from two directories (MEGA and DNZ) into a single combined Passim-friendly JSONL file.
- `run_passim.py`: Script to run Passim on a combined JSONL corpus to detect text reuse.

Notebooks:
- `run_passim_on_texts.ipynb`: Notebook (If not using scripts) to run `passim` on MEGA and DNZ datasets to get quote detection results.
- `display_passim_pairs.ipynb`: Notebook to display sentence pairs from `passim`'s output JSON (`align.json`).

Data:
- `texts`: Directory containing the original text and XML files.
- `texts_json_for_passim`: Directory for storing preprocessed, Passim-compatible text files.
- `passim_output`: Directory for storing results generated by Passim.

### Reviewer Checklist
- [x] Review the code in scripts and notebooks.
- [ ] Check if the `passim` results are valid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@
 ### Scripts
 
 - Add `parse_html` script for converting the manifesto html files to plain text for sentence corpus input (one-time use)
+- Add `convert_xml` script for converting Das Kapital XML files to plain text files (one-time use)
+- Add `convert_texts_to_passim_corpus` script for procssing plain text files of two corpus into one combined corpus for `passim`
+- Add `run_passim` script for running `passim` to detect sentence pairs on Das Kapital and DNZ
 
 ### Misc
 


laurejt:
> > @tanhaow what was this branch built from? The commit history looks quite long...
> 
> Initially, it was built from one of your experiments’ branches that used passim... Experiment 2, I believe.

Ah, that's why the commit history is extremely long. Once the 0.2 release is out; I think it may make more sense to create a new branch with 0.2 release / develop as the base. This shouldn't cause any issues since your changes do not rely on any of the code from the experiments branch, and instead are dependent on existing `remarx` code (for the TEI-XML parsing).

@rlskoeser Feel free to overrule me if there's a better way to deal with this.

scripts/convert_texts_to_passim_corpus.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # Process second directory
+        print(f"Processing {dir2_path}...")
+        for record in process_directory_to_passim(str(dir2_path), series2):
+            outf.write(json.dumps(record, ensure_ascii=False) + "\n")
+            total_records += 1
+
+    print("\nConversion completed!")
+    print(f"Total records: {total_records}")
+    print(f"Output file: {output_path.resolve()}")
```
laurejt: Why is `resolve` being called? This doesn't seem necessary for a `pathlib.Path`.

scripts/run_passim.py
```diff
@@ -0,0 +1,132 @@
...
+        result = subprocess.run(
+            cmd, cwd=Path.cwd(), env=env, capture_output=True, text=True
+        )
+
+        if result.returncode == 0:
+            print("Passim completed successfully!")
+            print(f"Output is saved in: {output_dir.resolve()}")
+        else:
+            print("Passim failed.")
+            print("Error message:", result.stderr)
```
laurejt: use f strings
```suggestion
            print("Error message: {result.stderr}")
```

uv.lock
laurejt: This shouldn't change without the pyproject.toml also changing!

scripts/convert_texts_to_passim_corpus.py
laurejt: This script does not run out of the box since its dependencies are missing from the `pyproject.toml`

laurejt (commented):
Thank you for your efforts in becoming familiar with passim. Note that I did not run and only cursorily reviewed the notebooks.

I'll leave it to @rlskoeser to compile a list of specific requested changes based on both of our reviews. But here are my high-level comments:

- If notebooks are used, they should be `marimo` notebooks. While `.ipynb` notebooks are useful for review, they do not follow our best practices. Additionally, these notebooks should depend on the functionality in the various scripts rather than copying code into the notebooks. Otherwise, there's a good chance of divergent behaviors between the scripts notebooks.
- This PR includes a lot of files that should not have been modified for this issue. Perhaps it's worth creating a new branch with the right base?
- The scripts and presumably the notebooks do not run since dependencies are missing from the `pyproject.toml`. If we plan to include any of these scripts/notebooks in the repo, we should make an optional dependency group with any dependencies exclusively used by this PR.
- The results CSV's name `alignment_sentence_pairs` does not reflect the content since the text reuse matches do not correspond to sentences.
- `convert_xml.py` should be updated so that it imports the relevant `remarx` classes/methods
-  `convert_texts_to_passim_corpus.py` runs, but does not seem worth preserving since it doesn't generalize well. For any future efforts with passim, I expect I'll write a quick script that takes a text directory and series id as inputs and produces a passim-friendly JSONL corpus (this would be a slightly modified version of corppa's [build_text_corpus](https://github.com/Princeton-CDH/corppa/blob/develop/src/corppa/utils/build_text_corpus.py). Additionally, it destructively modifies the input texts so which makes the passim character indices unreliable.
- `run_passim.py` seemed okay overall but will need future modification to be efficient. One thing I found troubling was that the script deletes the specified output directory if it exists; this seems like unexpected/unsafe behavior to me. I think it's worth reconciling the discrepancies between this script and the reference [corppa script](https://github.com/Princeton-CDH/corppa/blob/develop/src/corppa/poetry_detection/passim/run_passim.py) (see my comments for specifics).

Issue #134, opened 2025-08-04
As a user, I want my TEI-XML sentence corpus to include line numbers so that I can get back to the specific part of the MEGA text.


mnaydan:
## Acceptance Testing Notes

Run `uv run remarx-app` to launch the app. 
Download the TEI file of the [first volume of Das Kapital](https://drive.google.com/file/d/151cShKSuICAGvCfik8vtokL8TTTLrJgl/view?usp=drive_link) .
Like what we did in the previous test, follow the instructions on the `Corpus Builder` page to build a sentence corpus for Das Kapital (it should take at least 4 mins to finish).

- [x] Verify the output csv file now has a `line_number` column
- [x] Confirm that the line numbers on each page are in increasing order.

Go to the [web interface](https://telota.bbaw.de/mega/) that hosts the original MEGA content. Jump to some random pages of the first volume of Das Kapital (MEGA II.5 1-654).
- [x] Verify that the line numbers of sentences match what’s being said in the CSV file.

EBaring:
Looks good to me

bn2409-svg:
Also looks good to me!

PR #249, opened 2025-10-15
Feature/add line numbers to corpus

tanhaow:
**Associated Issue(s):** #134 #248 

### Changes in this PR
- Added `line_number` field to each sentence in the corpus (available for TEI XML documents only)
- Added `get_extra_metadata()` method in `TEIinput` class to pass line number info
- Added `get_extra_metadata()` hook method in `FileInput` base class, allowing `TEIinput` class to override and provide line number info

### Notes
- Line number is only available for TEI XML documents; plain text documents do not include the `line_number` field
- Line numbers is captured from the n attribute of the <lb> element in the TEI XML.

### Reviewer Checklist
- [x] Verify that corpus generated from TEI documents includes `line_number` field
- [x] Review the code

### DIFF
--- a/src/remarx/sentence/corpus/tei_input.py
+++ b/src/remarx/sentence/corpus/tei_input.py
@@ -4,18 +4,23 @@ goal of creating a sentence corpora with associated metadata
 from the TEI.
 """
 
+import logging
 import pathlib
 import re
 from collections import namedtuple
 from collections.abc import Generator
 from dataclasses import dataclass, field
 from functools import cached_property
+from timeit import default_timer as time
 from typing import Any, ClassVar, NamedTuple, Self
 
 from lxml.etree import XMLSyntaxError, _Element
 from neuxml import xmlmap
 
 from remarx.sentence.corpus.base_input import FileInput, SectionType
+
+logger = logging.getLogger(__name__)
+
 
 TEI_NAMESPACE = "http://www.tei-c.org/ns/1.0"
 
@@ -311,6 +316,7 @@ class TEIinput(FileInput):
         :returns: Generator with dictionaries of text content, with page number and section_type ("text" or "footnote").
         """
         # yield body text and footnotes content chunked by page with page number
+        start = time()
         for page in self.xml_doc.pages:
             body_text = page.get_body_text()
             if body_text:
@@ -329,6 +335,11 @@ class TEIinput(FileInput):
                     "section_type": SectionType.FOOTNOTE.value,
                     "line_number": footnote.line_number,
                 }
+
+        elapsed_time = time() - start
+        logger.info(
+            f"Processed {self.file_name} with {len(self.xml_doc.pages)} in {elapsed_time:.1f} seconds"
+        )
 
     def get_extra_metadata(
         self, chunk_info: dict[str, Any], char_idx: int, sentence: str


src/remarx/sentence/corpus/base_input.py
```diff
@@ -60,6 +60,16 @@ def get_text(self) -> Generator[dict[str, str]]:
...
+    def get_line_number(
+        self, chunk_info: dict[str, Any], char_idx: int, sentence: str
+    ) -> int | None:
+        """
+        Hook method for subclasses to override to provide line number for a sentence.
+
+        :returns: Line number (1-indexed) if available, None otherwise
+        """
+        return None
+
```
rlskoeser: Let's rename `get_extra_metadata` or similar. Method should return a `dict` of fields and values to include; default implementation returns `None` or empty dict, whichever seems easier.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -95,6 +95,20 @@ def get_page_footnotes(self) -> list[xmlmap.XmlObject]:
...
+        Get the line number for a given character position in the text by counting newlines.
+        """
+
+        line_number = 1  # default to line 1
+        for i in range(min(char_pos, len(text))):
+            if text[i] == "\n":
+                line_number += 1
+
+        return line_number
+
```
rlskoeser: When iterating over text in `get_body_text()`:
- create a dictionary of line indices to track line number & character offset
- keep track of character offset as text is collected; move whitespace cleaning to inner loop so character offset is accurate for the cleaned up text

Revise this method to take character index; return the line break number for the line just before the character index for the start of the sentence.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -37,6 +37,25 @@ class BaseTEIXmlObject(xmlmap.XmlObject):
...
 
+class TEIFootnote(BaseTEIXmlObject):
+    """XmlObject wrapper for footnotes with convenience accessors."""
+
+    first_line_break = xmlmap.StringField("(.//t:lb[@n])[1]/@n")
+    "Line number of the first TEI line break (`lb`) within this footnote"
+
+    def get_first_line_number(self) -> int:
+        """Return the first line number for this footnote."""
+        return int(self.first_line_break)
```
rlskoeser: `xmlmap` includes an integer field; let's simplify this code. Here's a suggestion:

```suggestion
    line_number = xmlmap.IntegerField("(./t:lb[1]/@n")
    "Line number where this footnote begins, based on first TEI line beginning (`lb`) within this note"
```

rlskoeser (commented):
I have some preliminary comments on the revised xml handling, haven't made my way through all the new logic or tests yet and haven't tested locally, but submitting so you can see these.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -95,11 +106,39 @@ def get_page_footnotes(self) -> list[xmlmap.XmlObject]:
...
     def get_body_text(self) -> str:
         """
         Extract body text content for this page, excluding footnotes and editorial content.
+        While collecting the text, build a mapping of character offsets to TEI line numbers.
         """
-        body_text_parts = []
+        body_text_parts: list[str] = []
+        line_number_by_offset: OrderedDict[int, int] = OrderedDict()
+        offsets: list[int] = []
+        char_offset = 0
```
rlskoeser: dict order by insertion order is guaranteed starting in python 3.7 and we're well past that, so no need to use ordered dict.

I also don't understand why you need to keep double track of the offsets. Please simplify to use the class instance variables throughout (instead of local and then copied to class instance), and just use the line number offset mapping.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -122,24 +161,57 @@ def get_body_text(self) -> str:
...
                 continue
 
-            body_text_parts.append(text)
+            raw_fragment = str(text)
+            cleaned_fragment = re.sub(r"\s*\n\s*", "\n", raw_fragment)
+            if not body_text_parts:
+                cleaned_fragment = cleaned_fragment.lstrip()
+
+            if not cleaned_fragment:
+                continue
```
rlskoeser: fragment means something different to me (probably because of geniza). Can we change to `raw_text` and `cleaned_text` ? 

Why call `lstrip()` here and `rstrip()` later? Wouldn't it be simpler to omit here and call `strip()` at the end as before? Is there something you're correcting for with this?

rlskoeser (changes requested):
Haven't tested the code locally yet but wanted to give some feedback on simplifying the code.   

Let's leave line numbers blank (rather than default 1) if we ever can't determine, so that it will be easier to track down and correct possible errors in the logic / output.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -95,11 +106,39 @@ def get_page_footnotes(self) -> list[xmlmap.XmlObject]:
...
+            self.get_body_text()
+
+        line_number_by_offset: OrderedDict[int, int] = getattr(
+            self, "_line_number_by_offset", OrderedDict()
+        )
+        offsets: list[int] = getattr(self, "_sorted_line_offsets", [])
+
+        # When there are no line breaks with line numbers, default to line 1
+        if not offsets:
+            return 1
```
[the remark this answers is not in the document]
tanhaow: I looked up the alto file - it never happened. Originally I didn't include it but then our code coverage check in CI falls. I had to add these two lines just order to the code coverage pass (But they will never be reached in reality).

src/remarx/sentence/corpus/tei_input.py
tanhaow: I’ll rename them to `raw_text` and `cleaned_text`.

Needed to call `lstrip()` here because when iterate `following::text()` after <pb n="17"/>, the first two text nodes are `\n`,  `\n ` before we reach the actual text “Der Reichthum…”. If we join everything and call .strip() at the end, those leading newlines will still add to the char_offset we’re recording

rlskoeser: oh, of course, the strip impacts the character offset, that makes sense. please add a comment noting this.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -95,11 +106,39 @@ def get_page_footnotes(self) -> list[xmlmap.XmlObject]:
...
+
+        line_number = line_number_by_offset[offsets[0]]
+        for offset in offsets:
+            if offset > char_pos:
+                break
+            line_number = line_number_by_offset[offset]
+        return line_number
```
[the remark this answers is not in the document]
rlskoeser: ah, right, my mistake - we're looking for the first line past as a way to identify the immediately preceding line.

src/remarx/sentence/corpus/tei_input.py
```diff
@@ -122,29 +178,54 @@ def get_body_text(self) -> str:
...
+                    if body_text_parts and not body_text_parts[-1].endswith("\n"):
+                        cleaned_text = f"\n{cleaned_text}"
+                elif char_offset > 0:
+                    # Preserve explicit blank lines introduced by <lb/> elements.
+                    cleaned_text = "\n"
+
+                line_attr = lb_element.get("n")
+                if line_attr is not None:
+                    self.line_number_by_offset[char_offset] = int(line_attr)
+                last_handled_lb = lb_element
```
rlskoeser: This seems complicated — I think I'm missing something about why this is needed. If we have mixed content immediately after the `<lb>` tag is there a case where there's no tail text to ensure we can add a newline? Or do we not hit the `<lb>` tag as a parent in the right sequence?

rlskoeser: Yes, I found the right example in your unit tests - it looks indented the way you've formatted it here, but I see in the xml fixture that's exactly the case, no tail text. Thanks for such good unit tests to cover these odd cases! Thinking about whether there's any easy way to simplify this...

tanhaow: I was able to find these cases because you suggested yielding undecided line numbers `None` instead of `1`. Otherwise, I would have missed them! They were indeed easy to overlook. But yeah, I also think the current logic looks complicated. Hope there's a way to simplify this

rlskoeser (approved):

Issue #256, opened 2025-10-15
Make Marimo Corpus Builder notebook accept zipfile input


tanhaow:
- Change the `Corpus Builder` Marimo notebook to make it accept zipfile input.

Issue #253, opened 2025-10-15
add validation method to check ALTO XML zipfile


rlskoeser:
- check that contents are xml files
- check that xml files are ALTO XML

Issue #250, opened 2025-10-15
create preliminary FileInput class for ALTO XML


rlskoeser:
New subclass for alto input
- takes a zip file as input
- field names include `section_type`
- preliminary`get_text` method : iterate over xml files in the zipfile (print out for now)

rlskoeser:
Here's some of the code in PPA where we're working with files content within a ZipFile, in case it's a helpful example:
https://github.com/Princeton-CDH/ppa-django/blob/0fb7fb9a50429ca26c26d686297cc9b4c521c6a2/ppa/archive/models.py#L1065-L1077

Also recommend looking at the [python zipfile documentation](https://docs.python.org/3/library/zipfile.html).

PR #258, opened 2025-10-17
Feature/fileinput class for alto

tanhaow:
**Associated Issue(s):** #250 #253 #256 

### Changes in this PR
- Added a preliminary FileInput class for ALTO XML that takes in a zipfile and iterates over its members
- Added a function to validate ALTO zipfiles and cache the discovered member filenames (to avoid rescanning)
- Added unit tests and a sample ALTO zipfile fixture

### Notes
- Used ALTO v4 namespace (aligned with what was used in [geniza](https://github.com/Princeton-CDH/geniza/blob/5d3e497acb5835346890206fde9e8037e438dff4/geniza/corpus/management/commands/escr_alto_to_annotation.py#L22))
- Checked the Marimo `Corpus Builder` notebook’s upload control: it draws allowed extensions straight from the backend input classes (`filetypes=FileInput.supported_types()`), and because our ALTOInput subclass now declares the `.zip` file type, the widget already accepts zip uploads. Also tested uploading a zip in the UI, which worked. So #256 can be closed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Sentence corpus creation
 
-- Add preliminary ALTO XML zipfile input with validation logic
+- Preliminary support for ALTO XML input as a zipfile of multiple pages
 
 ## [0.2.0] - 2025-10-15
 


tanhaow:
> I expect some of this logic to change a fair bit when we integrate the xml parsing, but I think it would be helpful to merge this in as an initial implementation to build on.
> 
> I'd like to keep the validation issue open because that's one of the pieces I expect to change most. We didn't discuss specifics, but I think we can be a bit more forgiving. I would propose the following:
> 
> * non-xml files are ignored (but logged - maybe info)
> * non-alto xml files are ignored (but log a warning)
> * error if no alto xml files are found in the zipfile
> 
> I also think we won't need to iterate over the files twice, the parsing should also handle the validation piece you're doing now - i.e., if we attempt to open a non-alto file as our alto xml object, then the object should raise an exception that we can catch and report here in the input class.

Thank you for suggesting this. I’ve added them as a note in the validation issue and will address them.

Issue #271, opened 2025-10-28
As a user, I want ALTO sentence corpora to include and filter based on block labels, so the output is more specific and type of content is labeled.


rlskoeser:
## testing notes
Check the output sentence corpus CSV file generated with ALTO XML zipfile input, and verify that:
- [x] resulting CSV has section information based on the block labels from the ALTO
- [x] CSV sections only include text, footnotes, and Title

It may be helpful to compare specific content in eScriptorium with the sentences that are generated to confirm that the block tags are being handled properly.

EBaring:
Sorry, can you point me to where we can find the alto zip files.

EBaring:
I know this is closed. But it worked for me. Just one thing. One one of the footnotes I looked at, the OCR got the order wrong. Look at line 204 on p. 9. The footnote picks up the author (here "Der Herausgeber" before the final bit of the sentence.

PR #273, opened 2025-10-29
Use ALTO tags for section information; set default list of types to include

rlskoeser:
**Associated Issue(s):** #271 

### Changes in this PR

- Add mapping for tag ids and labels in ALTO xml with property dictionary lookup
- Add simple access to block tag label when iterating over blocks in order
- Set a default filter of the block types to include
- Set block label as section type in sentence corpora

### Notes

- We might need to control the included types in future; this is partially implemented but without a mechanism at the command line or marimo app interface level

### Reviewer Checklist

- [ ] Review the code
- [ ] Test creating an ALTO sentence corpus; confirm it includes only the expected sections, section type is more informative

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 0.4.0
+
+### Sentence corpus creation
+
+- ALTO input now uses block-level tags for filtering and section type in sentence corpus
+    - By default, only includes blocks tagged as text, footnote, Title, or untagged
+
 ## [0.3.0] - 2025-10-27
 
 ### Sentence corpus creation


src/remarx/sentence/corpus/alto_input.py
```diff
@@ -146,6 +172,13 @@ class ALTOInput(FileInput):
     file_type: ClassVar[str] = ".zip"
     "Supported file extension for ALTO zipfiles (.zip)"
 
+    default_include: ClassVar[set[str]] = {"text", "footnote", "Title"}
+    "Default content sections to include"
+
+    filter_sections: bool = True
+    "Whether to filter text sections by block type"
+    # do we need a way to specify custom includes?
```
tanhaow: I think this can be a useful feature later for enhancement and evaluation, so we can try filtering out different tags to see if the corpus becomes better. So it's nice to have it

tanhaow (approved):
Looks good to me. For the question on if we want a way to specify custom includes, I think it's a nice-to-have but not a must-have (since Ed has stated what types he wants to include). But it might be useful for ourselves later in enhancement. So it's good to add it. Other than this, everything looks good to me!

rlskoeser:
> ... For the question on if we want a way to specify custom includes, I think it's a nice-to-have but not a must-have ... But it might be useful for ourselves later in enhancement. So it's good to add it. 

I don't have a clear vision for an easy way to pass in parameters right now - our sentence corpus creation inputs so far don't have any options that are specific to the input content-type. I think they may in the future but not sure how to handle it, so I'm going to merge this as is until we have a clear need to access this, because that will inform how we build it.  (For one-offs for ourselves, we can always modify the list or turn off the filtering manually in the code.)

Issue #269, opened 2025-10-27
improve TEI input handling


mnaydan:
## Testing notes

Use the app to create a sentence corpus from the MEGA TEI and review the resulting sentence corpus file to check the following:

- [x] output skips the title page (marked as 'editorialHead' in the xml, we are skipping it entirely)
- [x] output does not include tables (includes the table of content)
- [x] Revised: output does not include multi-line math formulas (e.g., fractions); single-line formulas are not tagged in xml and thereby cannot be filtered out
- [x] sentences for all body content come first followed by all footnote content
- [x] footnote content does not include footnote label
- [x] footnote sentences have the correct page number
- [x] body content does not include footnote references but does include text immediately after footnote references (inadvertently admitted previously)
- [x] sentences don't preserve line breaks and italicized text doesn't introduce line breaks
- [x] sentences that cross page boundaries are not broken and have correct page and line numbers (only works for sentences in a single paragraph in the TEI - continuing paragraphs seem to be broken when there are footnotes; an example of one that should work is bottom of page 26 / start of page 27)

## Development notes
- [x] skip math formulas (for now at least)
- [x] skip tables (for now, at least)
- [x] skip title page (`<div type="editorialHead">`)
- [x] Confirm table of contents is skipped by skipping tables; if not, skip based on div type (`<div2 type="inhaltsVZ"`)
- [x] omit labels for footnotes
- [x] bug: include tail text for footnote reference in body content (currently seems to be skipped)
- [x] generate text at the paragraph level (iterate paragraphs within a page) and don't preserve line breaks
- [x] ensure that inline markup (e.g. `<hi>` or `<ref>`) does not introduce line breaks

tanhaow:
@rlskoeser  Thanks for the pointer. Here's what I’m thinking about it:

Footnote labels: I like your idea. I’ll turn TEIFootnote.text into a list field to rebuild the footnote text from every descendant text node except those underneath`t:label[@type='footnote']`. This should drops the leading “1)” and handle tables in footnotes without string surgery. 

Math / fractions / tables: I think it's a good idea to adapt the `is_footnote_content` method - maybe call it `is_structural_content`. It should look for ancestors like m:math, table, list, figure, or the editorial/title divs. Both body paragraphs and footnotes can run through this helper method.

[...]

PR #277, opened 2025-10-30
Feature/improve tei handling

tanhaow:
**Associated Issue(s):** #269

### Changes in this PR
 Refactored TEI input handling logic to address problems pointed out in issue #269:
    - Body text is now yield at the paragraph level
    - Structural content (tables, math formulas, editorial divs) is filtered out in both body text and footnote text.
    - Footnote labels (e.g., "1)", "2)") are removed from note text (but stored separately in case we still need them).
    - Bug Fix: Inline footnote references now preserve trailing punctuation that follows `<ref type="footnote">`.

### Reviewer Checklist
- [ ] Review the code
Run TEI and check the output, check:
- [ ] Footnote labels are omitted
- [ ] Paragraphs that wrap pages correctly show `continued=True` label
- [ ] no more math formulas, tables, and title pages in the output (The output should look much cleaner now).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - ALTO input now uses block-level tags for filtering and section type in sentence corpus
     - By default, only includes blocks tagged as text, footnote, Title, or untagged
+- Refactored TEI input handling logic:
+    - Body text is now yield at the paragraph level
+    - Structural content (tables, math formulas, editorial divs) is filtered out in both body text and footnote text.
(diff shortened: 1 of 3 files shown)


rlskoeser (commented):
I've been trying to figure out what to do about your implementation - I couldn't figure out how to simplify this approach, so I went back to develop and retooled to operate at the paragraph / block text level instead, which is much more efficient than page boundaries. 

I'd be glad for talking through your approach and how you arrived at it if that would be helpful. And I can give you feedback on other aspects of the code too, if it would be useful - some of the unit tests were testing logic in what seemed like unexpected places; it's always best to test as close as possible to the method or attribute.

PR #284, opened 2025-11-05
Feature/display realtime logging in UI

tanhaow:
**Associated Issue(s):** #

### Changes in this PR

- [ ] Implemented the realtime logging view for the corpus builder notebook: Watch the active log file with `mo.watch.file` and drive a hidden `mo.ui.refresh` timer so the panel re-renders every second. 

### Notes

- We need users to **refresh the browser** to start a fresh session to keep the logging panel updating and displaying the realtime logging info. Because Marimo ties UI elements to the current browser session, refresh events continue to use the original session ID. If that session expires, the refresh timer fires with the stale ID and Marimo raises `ValueError: Invalid session id`. According to marimo's documentation bot and the a few related issues, it seems like there isn’t a programmatic workaround yet... I added UI text reminding users to refresh the browser after they kick off a build.

image

- Currently the display logging logic is in a separate python file. I'm unsure if we want to keep it separate so it's easy to reuse, or we want to merge it into corpus builder.

### Reviewer Checklist
- [ ] Review the code
- [ ] Test the marimo notebook to check if the logging works

### DIFF
--- a/tests/test_app/test_log_viewer.py
+++ b/tests/test_app/test_log_viewer.py
@@ -38,7 +38,6 @@ def test_read_log_tail_handles_missing_file(tmp_path: pathlib.Path) -> None:
         ("a\nb\nc", 2, "b\nc"),
         ("line", 5, "line"),
         ("line1\nline2\n", 1, "line2"),
-        ("line1\nline2", 0, ""),
     ],
 )
 def test_read_log_tail_returns_expected_lines(
@@ -88,4 +87,8 @@ def test_render_log_panel_handles_stdout_configuration() -> None:
         refresh_ticks=0,
     )
 
-    assert "Logging is configured to stdout" in panel._repr_html_()
+    html = panel._repr_html_()
+    assert (
+        "Logging is not configured to write to a file for this session; "
+        "no log file is available to preview."
+    ) in html


src/remarx/app/log_viewer.py
```diff
@@ -0,0 +1,91 @@
...
+
+    name: str
+
+    def read_text(
+        self,
+        encoding: str | None = None,
+        errors: str | None = None,
+    ) -> str:  # pragma: no cover - Protocol
+        """Return the file contents as text."""
+
```
rlskoeser: Is this purely for typing? Could you not use pathlib or the `FileState` object that marimo returns?

src/remarx/app/log_viewer.py
```diff
@@ -0,0 +1,91 @@
...
+        log_tail = read_log_tail(watched_log)
+        if log_tail is None:
+            display_text = (
+                f"Waiting for log file `{log_file_path.name}` to be created..."
+            )
+        else:
+            display_text = log_tail or "[no log messages yet]"
+    else:
+        display_text = (
+            "Logging is configured to stdout for this session; "
```
rlskoeser: Could be stdout or sdterr, let's just say not a file.

```suggestion
            "Logging is not configured to write to a file for this session; "
```

rlskoeser (commented):
I can't believe you got this to work! I wish we'd found a simpler solution, but this is definitely functional and usable. 

I had a thought about the session - I would guess there is a default session timeout that is configured either by fastapi or by uvicorn.  I found a mention of a timeout keep alive option for uvicorn, but I'm not convinced that's what we want: https://stackoverflow.com/a/78768373

It's worth looking a bit more to see if we can handle this with a config change rather than requiring a manual reload.

I've suggested some minor revisions and asked one question; once those are resolved and you've looked into the session timeout, I think we can merge this.

rlskoeser (approved):

PR #296, opened 2025-11-13
Update script & app with input options to control consolidation

### LINKED ISSUE #288: As a user, when I find quotes, I want them to be consolidated by default so that I can analyze at the passage level.
## Testing Notes:

Go to the `Quote Finder` notebook in the app. At the bottom of `2. Select Output Location` section, you should see a toggle button to control quote consolidation; 

- [ ] Confirm this toggle exists and it should be turned on by default.
- [ ] run the find quotes step both ways and confirm that they are consolidated when the toggle is on and not consolidated when they are not (easiest way to check is probably to look for the `num_sentences` field, which will only be present in the consolidated quotes)
- Run find quotes with at least two original sentence corpora included and consolidation turned on and review the output. It may be helpful to filter on the `num_sentences` field and exclude rows with value 1 (not consolidated)
   - [ ] only sentence pairs that are sequential in both original and reuse should be consolidated
   - [ ] number of sentences is reported accurately
   - [ ] combined metadata is accurate
   - [ ] no sentences are combined across multiple original sentence corpora

---

Technical criteria:

- [x] in app, checkbox to control selection
- [x] default is consolidate
- [x] in CLI, use a flag

Acceptance criteria:

- [ ] user can choose to see consolidated/unconsolidated outputs

rlskoeser:
**Associated Issue(s):** resolves #288

### Changes in this PR

- Update `find_quotes` script to add a command line option to control consolidation
- Update find quotes notebook with a toggle to control consolidation
- Refactor `find_quotes` script to remove unnecessary wrapper method and use patch in unit tests instead of monkeypatch

### Notes

- When I was testing in marimo, I ran find quotes once with the toggle on and then changed the selection and ran again and it still consolidated - I'm not sure if I did something wrong, or if there is some state / reactivity that I'm not missing. The value is correctly displayed.

### Reviewer Checklist

- [x] Review code changes
- [x] Test CLI parameter
- [x] Test notebook parameter; test if it successfully uses the new option when the toggle changes after running

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,16 @@
 
 ## 0.5.0.dev
 
+### Sentence corpus creation
+
 - ALTO-generated corpora now include per-sentence `title` and `author` metadata derived from ALTO block tags
+
+### Quotation detection
+
+- Add consolidation logic to group quotes that are sequential in both original and reuse texts
+    - Quote consolidation is enabled by default but configurable:
+        - `remarx-find-quotes` script has a boolean parameter (`--no-consolidate`)
+        - App now has a toggle to control consolidation; enabled by default
 
 ## 0.4.0
 
@@ -25,7 +34,6 @@
 ### Quotation detection
 
 - Add the `remarx-find-quotes` CLI script to run the quote finder, with a `--benchmark` flag to collect performance metrics
-- Add consolidation logic to group quotes that are sequential in both original and reuse texts
 
 ### Application
 


rlskoeser:
> I didn't meet the problem you mentioned in notes about changing the selection and ran again, and it still consolidated. I was able to generate unconsolidated output once I toggled it off.

Oh good!  Thanks for testing - I was hoping that behavior was a fluke (or user error).

Issue #300, opened 2025-11-14
update find quotes script to support specifying multiple original files


rlskoeser:
## dev notes
- [x] for original corpus input : support a directory in addition to single file; if directory, use all csv files in that directory
- [x] if not specified, use the same default path as the app for original corpus 
- [x] call find quotes method with multiple original corpora files when specified

rlskoeser:
We don't expect project team members do acceptance testing on the command-line scripts; @tanhaow and I have both tested this functionality and it is working for us, so I'm going to close the issue.

PR #305, opened 2025-11-18
Refactor pairs find quotes methods to take multiple original files as input

### LINKED ISSUE #301: refactor methods in pairs module to support multiple input files for original corpora
### dev notes 

requirements for refactor:
- `find_quote_pairs` needs to take multiple original files
- separate out logic for generating embeddings for original and reuse files 
- method that generates embeddings needs access to file for future caching

rlskoeser:
**Associated Issue(s):** resolves #301

### Changes in this PR

- Rename `load_sent_df` to `load_sent_corpus` and shift logic for generating embeddings to that method
- Revise `find_quote_pairs` method to take a list of original corpus filenames
- Update find quote pairs command line script to pass original file as a list (full support for multiple will be added in #300 )
- Revise and simplify benchmarking logic so the logic is all contained in one method; add/revise more stats
- Updated existing tests to work with the revised code
- Added a multifile variant of the integration test to check the new logic

### Notes

- Revised benchmark stats to include corpus sizes (orig, reuse, pairs); dropped total time, since we can easily calculate in the spreadsheet
- Some variations to current functionality seem like they might be nice, but decided they should be out of scope until/unless we need them. Noting here as possibilities for future:
   -  find quotes method returned a dataframe instead of or in addition to writing out to a file
   - load sentence corpus might not always require embeddings, but we don't currently need it without and it was nicer to do them together

### Reviewer Checklist

- [x] Review code changes with an eye towards upcoming functionality (multifile support, caching embeddings) 
- [x] Test running locally - command line script should work as previously

### DIFF
--- a/src/remarx/quotation/pairs.py
+++ b/src/remarx/quotation/pairs.py
@@ -151,11 +151,11 @@ def compile_quote_pairs(
     detected_pairs: pl.DataFrame,
 ) -> pl.DataFrame:
     """
-    1Combine sentence metadata from original and reuse corpora with detected
+    Combine sentence metadata from original and reuse corpora with detected
     sentence pair identifiers to form quote pairs. The original and reuse
     corpus dataframes must contain a row index column named `original_index` and
     `reuse_index` respectively. Ideally, these dataframes should be built using
-    [load_sent_df][remarx.quotation.pairs.load_sent_df].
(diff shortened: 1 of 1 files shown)


rlskoeser:
@tanhaow it seems easiest to restore the functionality we have, so I'm going to do that. I'll also update the app to work with the refactored find quotes method.

PR #323, opened 2025-12-03
Revise changelog check to only run on open pull requests

### LINKED ISSUE #320: fix changelog check to not run on closed PR when this sprint label is removed

rlskoeser:
**Associated Issue(s):** resolves #320 

### Changes in this PR

- Add a conditional to check that pull request state is open

### Notes

- To resolve case where the workflow runs and fails when labels are removed from closed PRs

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     # only run on open pull requests  (do not run if a label changes
     # on a closed PR, e.g., when "this sprint" label is removed)
-    if: ${{ github.event.pull_request.state != "closed" }}
+    if: github.event.pull_request.state == 'open'
     steps:
       - name: Check changelog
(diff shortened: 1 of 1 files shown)


rlskoeser:
@tanhaow the check isn't running but it should be - something is still wrong with my conditional test.

tanhaow:
> @tanhaow the check isn't running but it should be - something is still wrong with my conditional test.

Ah, got it. I'll take a look into it now.

rlskoeser:
> @rlskoeser how can we possibly test it to see if it runs on closed PR without merging to develop first?

I was thinking:
- close the pr without merging
- add a label
- remove a label
- reopen
- add / remove labels if we still need to check

rlskoeser:
It worked! When I closed it, the job shows up in the action queue, but then reports that it was skipped.

https://github.com/Princeton-CDH/remarx/actions/runs/19901140623

Screenshot 2025-12-03 at 11 29 59 AM

PR #335, opened 2026-01-08
Feature/rejoin hyphenated words in ALTO

### LINKED ISSUE #279: ALTO body content correctly re-joins end-of-line hyphenated words
### Acceptance Testing

Run `uv run remarx-app` to launch the app. 
Follow the instructions on the `Sentence Corpus Builder` page to generate an ALTO corpus (like using `DNZ vol 1896-97a`, etc.).

Open the output `csv` file:

- [x] Verify that end-of-line words split by double oblique hyphen `⸗` are now rejoined 
       example: `europäisch⸗demokratischen` --> `europäischdemokratischen`
- [x] Verify that end-of-line words split by ASCII hyphen `-` are now rejoined
       example: `Gewerkschafts-bewegung` --> `Gewerkschaftsbewegung`

This optimization improves corpus quality for embeddings and matching.

---

### Dev Notes

Ed mentioned this is a special character

tanhaow:
**Associated Issue(s):** resolves #279

### Changes in this PR

- Added text preprocessing in `ALTOInput.get_text()` to rejoin end-of-line hyphenated words
- Added `test_alto_text_cleaning()` with fixture to verify the cleaning works

### Notes
- I carefully reviewed both TEI and ALTO corpus again. There are multiple types of hyphens and dashes in TEI, but luckily for ALTO we only have ASCII hyphen. So only ASCII hyphen-minus (`-`) is handled and rejoined.
- I overthought about this issue at first (wrote a complicated logic on another branch)... still not sure if the current solution is what we want. But now the end-of-line hyphenated words are indeed rejoined. Look forward to your suggestions.
- Created separate test fixture to avoid breaking existing tests

### Reviewer Checklist
- [x] Verify ALTO processing rejoins hyphenated words like `"Geschichts-\nauffassung"` → `"Geschichtsauffassung"`
- [x] Review the code

### DIFF
--- a/src/remarx/sentence/corpus/alto_input.py
+++ b/src/remarx/sentence/corpus/alto_input.py
@@ -242,7 +242,7 @@ class ALTOInput(FileInput):
 
                     block_text = block.text_content
                     # Clean up hyphenated line breaks from ALTO physical layout
-                    # Rejoin words split by ASCII hyphen (-) or two-em dash (⸗) followed by newline
+                    # Rejoin words split by ASCII hyphen (-) or double oblique hyphen (⸗) followed by newline
                     block_text = re.sub(r"[⸗-]\n", "", block_text)
                     chunk = {
                         "text": block_text,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 - Changed the sentence tokenizer from Stanza to spaCy, with default model `de_core_news_sm`.
 - Added `pip` as a dependency so spaCy models can be downloaded automatically when missing.
 - Added sentence filtering to exclude 1-2 word sentences and punctuation-only sentences from corpora to improve quote detection quality.
-- Rejoin end-of-line hyphenated words (ASCII hyphen-minus and two-em dash `⸗`) in ALTO text extraction to improve corpus quality for embeddings and matching.
+- Rejoin end-of-line hyphenated words (ASCII hyphen-minus `-` and double oblique hyphen `⸗`) in ALTO text extraction to improve corpus quality for embeddings and matching.
 
 ## 0.5.0
 


src/remarx/sentence/corpus/alto_input.py
```diff
@@ -240,6 +240,8 @@ def get_text(self) -> Generator[dict[str, str], None, None]:
                         collected_metadata = True
 
                     block_text = block.text_content
+                    # Clean up hyphenated line breaks from ALTO physical layout
+                    block_text = block_text.replace("-\n", "")
```
rlskoeser: I thought this was the character we are expecting for hyphenated line breaks in the ALTO: ⸗

tests/test_sentence/test_corpus/test_alto_input.py
```diff
@@ -469,6 +470,32 @@ def test_altoinput_error_empty_zip(tmp_path: pathlib.Path):
...
+    )
+
+    # Should find the rejoined text in the output
+    assert any("Geschichtsauffassung" in text for text in chunk_texts), (
+        "Geschichtsauffassung should be properly rejoined"
+    )
+    assert any("Socialdemocrat" in text for text in chunk_texts), (
+        "Socialdemocrat should be properly rejoined"
+    )
+
```
rlskoeser: Would be good to add 1-2 negative tests of content that you do not expect to be modified.

src/remarx/sentence/corpus/alto_input.py
rlskoeser: Would it make more sense for the cleanup to happen in the `text_content` property on the `TextBlock` class? I'm trying to think what scenarios we might still want access to the raw text. Maybe two variants, `raw_text_content` and then a `text_content` that does whatever cleaning steps we want to apply.

We might also want to collapse newlines after this cleanup is done (IIRC we did something like that on the TEI).

But it could be that this is more specific to our alto input class than the xml handling, in which case keeping it where you have it now seems best.

tanhaow: On the older branch, I handled multiple hyphens, including this one. I guess I was a bit uncertain and worried I overcomplicated things. There are indeed cases of this normal hyphen (-) at end-of-line words. I will add this hyphen (⸗) back.

tests/test_sentence/test_corpus/test_alto_input.py
```diff
@@ -20,6 +20,7 @@
 
 FIXTURE_DIR = pathlib.Path(__file__).parent / "fixtures"
 FIXTURE_ALTO_ZIPFILE = FIXTURE_DIR / "alto_sample.zip"
+FIXTURE_ALTO_HYPHENATED_ZIPFILE = FIXTURE_DIR / "alto_hyphenated_test.zip"
 FIXTURE_ALTO_PAGE = FIXTURE_DIR / "alto_page.xml"
 FIXTURE_ALTO_PAGE_WITH_FOOTNOTES = FIXTURE_DIR / "alto_page_with_footnote.xml"
```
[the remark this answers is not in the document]
rlskoeser: I see one example in alto_page.xml : `Gewerkschafts-`

But you could always introduce an example into the existing fixture as long as it doesn't break the other unit tests.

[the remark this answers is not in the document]
tanhaow: Initially I tried introduce two examples into the existing fixture but it broke too many existing unit tests, so I reverted it... But thank you so much for finding that example. Can't believe I missed it. I'll modify it to just use the fixture we have.

src/remarx/sentence/corpus/alto_input.py
rlskoeser: I think this one ⸗ is more common, but I'm not sure about the others or if anything has changed with the OCR output. This is why I recommended checking with Ed to confirm.  (Our notes indicate they intended to use a specific character for end-of-line hyphenated words.)

However, handling multiple characters doesn't require much more complexity than what you have in this branch: instead of a text replace, you can just switch to a regex replace that allows for multiple possible hyphen characters occurring just before a newline (maybe something like `r"[⸗-]\n"` ? ), and replace all occurrences.

rlskoeser: If you can get the regex replace all with these two characters that's probably good enough for this phase. Once we're using a regex with a set of characters it shouldn't be too hard to change the set of characters we're looking for.

rlskoeser (approved):
🚀

PR #321, opened 2025-12-01
Update find quotes script to support multiple original files and default corpus directories

tanhaow:
**Associated Issue(s):** resolves #300 

### Changes in this PR
- remarx-find-quotes now accepts either a single original-corpus CSV or a directory of CSVs.

### Notes
- If no original path is provided, it falls back to `~/remarx-data/corpora/original`, matching the app default.

### Reviewer Checklist
- [x] Test `remarx-find-quotes` script with a single CSV, a directory, and no specified original path (default) scenarios to confirm path handling

PR #361, opened 2026-05-11
Fix original_index misalignment when using multiple original corpus f…

tanhaow:
**Associated Issue(s):** resolves #362

### Bug

When multiple original corpus files were concatenated, each file's row index restarted from 0, creating duplicate values in `original_index`. The Voyager search returns globally unique positions (0 to N-1 across all files), so the join was matching against the wrong rows — producing incorrectly matched quote pairs.

There are two failure modes:
- If the global position falls within the first file's index range (0 to N-1), the join matches against the wrong sentence in the first file — producing a false positive
- If the global position exceeds the first file's range, the inner join silently drops the row — producing a false negative with no error or warning

In practice, results are heavily skewed toward the first original file passed. Matches from all other files are either dropped or mismatched.

Introduced in commit d89fa82 (#321).

### Changes in this PR

- Fix `original_index` misalignment: reassign as a globally unique range after `pl.concat`
- Update `test_find_quote_pairs_integration_multifile` to place the matched sentence in the second file, so its global index exceeds the first file's range — the scenario that was silently failing before the fix
- Update changelog

### Notes

- Bug only affects runs with multiple original files; single-file runs are unaffected
- All previously generated CSVs from multi-file runs should be considered invalid and re-run
- This slipped through because the existing multifile integration test used data small enough that the matched sentence's index happened to fall within the first file's range — the collision was never triggered
- The research team's review also did not catch it, partly because the output still contained many correct matches from the first file

### Reviewer Checklist

- [x] Verify fix logic in `pairs.py` — confirm `original_index` is reassigned after concat
- [x] Re-run quote detection with multiple original files and spot-check that matched pairs are semantically correct

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## [develop]
+
+### Bug fixes
+
+- Fixed a bug where using multiple original corpus files produced incorrect quote pairs. When files were concatenated, each file's row index restarted from 0, creating duplicate values in `original_index`. The Voyager search returns globally unique positions, so the join matched against the wrong rows. Fixed by reassigning `original_index` as a globally unique range after concatenation. Introduced in #321, fixed in #362.
+
 ## [1.0.1] - 2026-01-20
 
 - Updated technical design document to reflect 1.0 functionality
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,7 @@
 
 ## [1.0.2]
 
-### Bug fixes
-
-- Fixed a bug where using multiple original corpus files produced incorrect quote pairs (#362)
+- Bug fix: correct indexing for quotation detection when multiple original corpus files are selected as input ([#362](https://github.com/Princeton-CDH/remarx/issues/362))
 
 ## [1.0.1] - 2026-01-20
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 # CHANGELOG
 
-## [develop]
+## [1.0.2]
 
 ### Bug fixes
 
-- Fixed a bug where using multiple original corpus files produced incorrect quote pairs. When files were concatenated, each file's row index restarted from 0, creating duplicate values in `original_index`. The Voyager search returns globally unique positions, so the join matched against the wrong rows. Fixed by reassigning `original_index` as a globally unique range after concatenation. Introduced in #321, fixed in #362.
+- Fixed a bug where using multiple original corpus files produced incorrect quote pairs. When files were concatenated, each file's row index restarted from 0, creating duplicate values in `original_index`. The Voyager search returns globally unique positions, so the join matched against the wrong rows. Fixed by reassigning `original_index` as a globally unique range after concatenation.
 
 ## [1.0.1] - 2026-01-20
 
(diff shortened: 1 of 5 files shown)


src/remarx/quotation/pairs.py
```diff
@@ -204,7 +204,11 @@ def find_quote_pairs(
         original_vecs.append(vecs)
     # combine dataframes and vectors, preserving order
     # use diagonal concat method, since fields may vary by input type
-    original_df = pl.concat(original_dfs, how="diagonal")
+    # replace per-file index with a globally unique one across all files
+    combined_df = pl.concat(original_dfs, how="diagonal")
+    original_df = combined_df.with_columns(
+        pl.int_range(pl.len(), dtype=pl.UInt32).alias("original_index")
+    )
```
rlskoeser, line 212: I think this can be done in a single line and use the `with_row_index` method. Something like this (comments will need to be updated & harmonized):

```suggestion
    # replace per-file index with a globally unique one across all files
    original_df = pl.concat(original_dfs, how="diagonal").with_row_index("original_index")
```

tests/test_quotation/test_pairs.py
```diff
@@ -253,7 +253,15 @@ def test_find_quote_pairs(
...
-    mock_compile_pairs.assert_called_once_with(orig_df, reuse_df, ["sent_pairs"])
+    # orig_df passed to compile_quote_pairs has original_index reassigned to be
+    # globally unique; check that the other columns match and index is correct
+    actual_orig_df = mock_compile_pairs.call_args.args[0]
+    assert (
+        actual_orig_df["original_text"].to_list() == orig_df["original_text"].to_list()
+    )
+    assert actual_orig_df["original_index"].to_list() == [0, 1]
+    assert mock_compile_pairs.call_args.args[1].equals(reuse_df)
+    assert mock_compile_pairs.call_args.args[2] == ["sent_pairs"]
```
rlskoeser, line 266: This is testing your fix but I don't think it is actually testing the bug. Can we come up with a minimal test case that fails without your fix?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # CHANGELOG
 
+## [1.0.2]
+
+### Bug fixes
+
+- Fixed a bug where using multiple original corpus files produced incorrect quote pairs. When files were concatenated, each file's row index restarted from 0, creating duplicate values in `original_index`. The Voyager search returns globally unique positions, so the join matched against the wrong rows. Fixed by reassigning `original_index` as a globally unique range after concatenation.
```
rlskoeser: I think this is too long and too implementation-specific; I'll think about how to revise as I test the change locally.

rlskoeser (commented):
Can you suggest the best way to test this so I can see the behavior for myself before and after your fix? I don't mind manufacturing small test examples if that's the easiest way to see what's happening.

I'm a little puzzled why this didn't result in errors - if Voyager returns indexes from any text beyond the first one, do we simply ignore it because there's no matching index in our combined dataframe?  (I expect I'll understand better after I've reminded myself how we implemented this and tested it locally.)

rlskoeser (approved):
I'm ready to sign off on this now.  As part of my comparison testing, I ran a small subset of input without the bugfix and I can see that the result set is _much_ smaller than the full set (due to the missing join indices) but also seems to pull somewhat randomly from different original files (which I guess must be due to polars join on non-unique field). 

I made a few small tweaks; the integration test now checks with input files in both order, and I added a warning to the readme. What do you think of the warning?  (We may need to revise depending on what version we release.)

PR #364, opened 2026-05-11
Feature/visualize notebook

tanhaow:
**Associated Issue(s):** resolves #363 

### Changes in this PR
- Create a Marimo notebook that generates the four embedding visualization figures used in presentations.

### Notes
- Fig 1: Matched pairs in 2-D t-SNE space
- Fig 2: Two corpora in the same semantic space
- Fig 3: Similarity bar chart (nearest-neighbor search)
- Fig 4: Pair neighbourhood

### Reviewer Checklist
- [ ] review the code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,7 @@ ignore = []
 # for test files, don't require docstrings or return type annotations
 "tests/**.py" = ["D", "ANN"]
 # marimo notebook: @app.cell wrapper functions can't carry type annotations
-"visualize_embeddings_notebook.py" = ["ANN"]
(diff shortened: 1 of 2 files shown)


rlskoeser (commented):
Thanks for sharing this, it's been helpful to look at and I hope to draw on it more when I get further into the comparison testing and deepen my own understanding of how remarx works with the embedding space and ANN.

As discussed, this isn't really set up or intended to be merged. Perhaps in future we could use private GitHub gists to share code like this.

(I figured I should review so we'd stop getting reminders.)

PR #378, opened 2026-07-20
Hotfix v1.0.3

rlskoeser:
**Associated Issue(s):** #381 

### Changes in this PR

- Security updates flagged by dependabot / GitHub security review. This update should resolve all the critical and high severity alerts on this repo.

### Notes

- I created a hotfix branch and then cherry-picked commits from the dependabot PRs, since they include potentially useful details.
- I tested locally after upgrading transformers and sentence-transformers and compared find quotes before and after the upgrade - the embeddings and quote output both looked good.


pyproject.toml
```diff
@@ -23,14 +23,14 @@ classifiers = [
...
+  "marimo>=0.23.9",
   "neuxml>=1",
   "numpy",
   "polars>=0.20.4",
   "ruff",
-  "sentence-transformers",
+  "sentence-transformers>=5.5.0",
   "spacy",
-  "transformers[torch]",
+  "transformers[torch]>=5.5.0",
```
laurejt: To include the minor update that gets rid of spurious warnings for a wide range of model architectures. See top level comment.

```suggestion
  "transformers[torch]>=5.6.0",
```

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
 
+## [1.0.3]
+
+Maintenance release. Updates to python dependencies flagged by security scans.
+
+## Direct dependencies (python deps + uv.lock)
+
+- Bump marimo minimum version to 0.23.9
+- Bump transformers from 4.56.2 to 5.5.0
+- Bump sentence-transformers minimum version to 5.5.0
```
laurejt:
```suggestion
- Bump transformers from 4.56.2 to 5.6.0
- Bump sentence-transformers minimum version to 5.5.0
```

laurejt (commented):
1. When running the find quotes portion of the app, I receive the following warning when the model is being loaded:
```
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 199/199 [00:00<00:00, 55909.07it/s]
XLMRobertaModel LOAD REPORT from: sentence-transformers/paraphrase-multilingual-mpnet-base-v2
Key                     | Status     |  |
------------------------+------------+--+-
embeddings.position_ids | UNEXPECTED |  |

Notes:
- UNEXPECTED:	can be ignored when loading from different task/architecture; not ok if you expect identical arch.
```

This appears to be an issue introduced in `transformers` 5.5. Based on this [issue](https://github.com/huggingface/transformers/pull/45437), this issue is benign but the issue is fixed in v5.6.0, so I suggest we change our lower bound to 5.6.

2. Just want to note that running `uv sync --resolution lowest` and `uv sync --resolution lowest-direct` both fail.

laurejt (approved):
Once the `transformers` version lower bound is bumped to 5.6.0, I think this is ready to go.

